PR #3, opened 2012-02-09
Fixes in float parsing code

nijel:
Please merge these changes, which make Little-CMS actually work with CCMX files on locales not using . as decimals separator.


andresemoreno:
[_cvrvbjh_]()

PR #17, opened 2013-05-09
cmake support

est77:
This pull request adds support for building Little-CMS using cmake.
Its tested on windows 7 and OSX 10.8.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -125,3 +125,6 @@ Added some checks for non-happy path, mostly failing mallocs
 -----------------------
 2.5 Maintenance release
 -----------------------
+
+Fixed a double free in recovering from a previous error in default intent handler.
+
--- a/src/cmscnvrt.c
+++ b/src/cmscnvrt.c
@@ -571,7 +571,9 @@ cmsPipeline* DefaultICCintents(cmsContext       ContextID,
         // Concatenate to the output LUT
         if (!cmsPipelineCat(Result, Lut))
             goto Error;
+
         cmsPipelineFree(Lut);
+        Lut = NULL;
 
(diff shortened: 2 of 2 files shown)


danielkaneider:
Any new thoughts about using CMake instead of all the current project files? It would be a replacement for all MS build files. It's not that the community would take care about changes, as it can be seen in this PR. Nowadays most C++ projects have CMake build files

emmenlau:
Thanks @mm2 and all for this nice library!

I just wanted to bring up this topic of cmake again. I can see that in the past it was not considered because it duplicates autotools. While I can very much relate to this issue, I also wanted to raise the concern that supporting autotools becomes harder and harder. We find basically zero people that know autotools, and I'm the last in the company that knows them (a bit). Literally everybody switched to cmake. Actually, the Microsoft-supported cmake-based C++ package manager `vcpkg` even maintains a port of a cmake-based lcms2 build in https://github.com/microsoft/vcpkg/tree/master/ports/lcms.

Would it be possible to reconsider cmake support? It must not support every option, and I can see that it should be community-supported...

mm2:
Hello,
Thanks! I think actually cmake is worth of some attention. Now is very mature and stable, and as you well say, autotools has become such mess that is very hard to maintain. I lately was stuck by the fact there is no way to specify a degree of optimization, like -Ofast in gcc. Sure, it is widely portable, but you have almost no control. So, after release 2.12 I will be glad to try to port your build scripts to upstream. Many thanks for that.

emmenlau:
That would be awesome! Lets maybe re-open this issue then, to track the status in the future? Ping me to help with cmake integration...

PR #40, opened 2015-03-24
Allow ptr alignment to be set at build time

chris-liddell:
This came up during testing of the second Ghostscript 9.16 release candidate, causing a bus error
on SPARC/Solaris with gcc.

At least some configurations of gcc on UltraSPARC require 8 byte alignment of
memory pointers (even on 32 bit user space, where sizeof(void *) == 4).

Rather than contaminate the source, provide ability to set the ptr alignment
on the compiler command line (on Unix, using CFLAGS when calling configure).


mm2:
Great! thanks for your contribution

chris-liddell:
On 24/03/15 13:46, Marti Maria wrote:
> Great! thanks for your contribution
> 

Thanks for pulling so quickly - makes life in the Ghostscript world so
much easier!

Chris

PR #47, opened 2015-05-28
Fix possible null pointer dereference

ya1gaurav:
hProfile is being checked against Null at line no. 716, which means it could be Null. But it is not checked against Null before line no. 711 & it is dereferenced in PrintProfileInformation.


mm2:
Alreaddy fixed in 2.8, thank you

Issue #48, opened 2015-06-02
cmsio0.c compile error (MSVC 2013)


torokze:
Hello,

I have a compile error in **cmsio0.c** at line **1857**.

**error C2440: '=' : cannot convert from 'int' to 'cmsTagSignature'**

the solution to cast 0 to cmsTagSignature:

**Icc->TagNames[i] = (cmsTagSignature)0;**

Regards
Zolee

mm2:
I cannot reproduce this anymore

PR #49, opened 2015-06-05
Fix typecast for cmsTagSignature

ya1gaurav:
Fix typecast problem in reference to https://github.com/mm2/Little-CMS/issues/48


mm2:

PR #50, opened 2015-06-05
Fix Null check.

ya1gaurav:
cmsStageAllocMatrix can return NULL, which can cause crash when dereference Multmat in next line. To avoid crash added Null check.
Also made a small change in documentation.


mm2:

Issue #43, opened 2015-04-30
transicc stack buffer overflow in TakeFloatValues > GetLine


mikispag:
There is a very simple stack buffer overflow in transicc:

```
$ ./transicc -i ../../testbed/test1.icc 
LittleCMS ColorSpace conversion calculator - 4.3 [LittleCMS 2.07]

Enter values, 'q' to quit
C? 1
=================================================================
==10338==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffc0e6606f0 at pc 0x7fa6864c31d2 bp 0x7ffc0e660060 sp 0x7ffc0e660018
WRITE of size 4096 at 0x7ffc0e6606f0 thread T0
    #0 0x7fa6864c31d1 in scanf_common ../../.././libsanitizer/sanitizer_common/sanitizer_common_interceptors_scanf.inc:307
    #1 0x7fa6864c3911 in __interceptor___isoc99_vscanf ../../.././libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:598
    #2 0x7fa6864c39f7 in __interceptor___isoc99_scanf ../../.././libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:630
    #3 0x40f38e in GetLine /home/mikispag/Downloads/lcms2-2.7/utils/transicc/transicc.c:630
    #4 0x40f994 in TakeFloatValues /home/mikispag/Downloads/lcms2-2.7/utils/transicc/transicc.c:746
    #5 0x40b55d in main /home/mikispag/Downloads/lcms2-2.7/utils/transicc/transicc.c:1274
    #6 0x7fa685bd7ec4 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21ec4)
    #7 0x40e52c (/home/mikispag/Downloads/lcms2-2.7/utils/transicc/transicc+0x40e52c)

Address 0x7ffc0e6606f0 is located in stack of thread T0 at offset 928 in frame
    #0 0x40f7af in TakeFloatValues /home/mikispag/Downloads/lcms2-2.7/utils/transicc/transicc.c:719

  This frame has 7 object(s):
    [32, 34) 'index'
    [96, 136) 'Name'
    [192, 232) 'Prefix'
    [288, 328) 'Suffix'
    [384, 640) 'ChannelName'
    [672, 928) 'Buffer'
    [960, 5056) 'Buffer' <== Memory access at offset 928 partially underflows this variable
[...]

PR #52, opened 2015-06-05
Fix stack buffer overflow in transicc

ya1gaurav:
With reference to https://github.com/mm2/Little-CMS/issues/43
Address sanitizer reported stack buffer overflow.
As Getline function reads 4095, in this case Buffer is of length cmsMAX_PATH(256).
Its length should be 4096.


mm2:
Already fixed

ya1gaurav:
buffer size fixed ?

Issue #20, opened 2013-07-11
out of bounds stack buffer read


251:
Type_NamedColor_Read() creates a "Root" buffer of size 33 without 0-termination (local buffer is uninitalised) and reads 32 bytes from io, also without check for null termination:

``` C
        char Root[33];

        memset(Colorant, 0, sizeof(Colorant));
        if (io -> Read(io, Root, 32, 1) != 1) return NULL;
        if (!_cmsReadUInt16Array(io, 3, PCS)) goto Error;
        if (!_cmsReadUInt16Array(io, nDeviceCoords, Colorant)) goto Error;

        if (!cmsAppendNamedColor(v, Root, PCS, Colorant)) goto Error;
```

cmsAppendNamedColor then applies strncpy with huge cmsMAX_PATH:

``` C
    if (Name != NULL) {

        strncpy(NamedColorList ->List[NamedColorList ->nColors].Name, Name, cmsMAX_PATH-1);
        NamedColorList ->List[NamedColorList ->nColors].Name[cmsMAX_PATH-1] = 0;
```

and strncpy() reads behind Root/Name. You may reproduce the bug as follows:

``` sh
>./configure CC="clang" CFLAGS="-g -O0 -fsanitize=address -fno-omit-frame-pointer -fPIC"
>make
>./utils/psicc/psicc -imin.icc
==24151== ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7fffc1104c24 at pc 0x4130c5 bp 0x7fffc1104270 sp 0x7fffc1103a28
READ of size 1 at 0x7fffc1104c24 thread T0
...
```

base64 representation of min.icc is

```
AAABCgAAAAAAAAAAbm1jbAAAAAAAAAAAAAAAAAAAAAAAAAAAYWNzcAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAABbmNsMgAAAJAAAAB6bmNsMgAAAAAAAAAAAAAAAQAAAABhYWFhYWFh
YWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFh
[...]

PR #53, opened 2015-06-05
Out of bound stack buffer read fixed.

ya1gaurav:
With reference to https://github.com/mm2/Little-CMS/issues/20
Root is not Null terminated and hence show undesired behaviour when passed in cmsAppendNamedColor and strncpy is done.
Added null termination for Root.
P.S -> I just added line no 3107, It is showing other changes by itself.


mm2:
Alteady fixed in 2.8, thank you

PR #54, opened 2015-06-05
Null pointer dereference fixed.

ya1gaurav:
GetDataFormat can return Null, in that case cmsstrcasecmp will dereference fld. Hence check to validate fld is added.


mm2:

PR #56, opened 2015-06-05
Fix multiple Null terminations.

ya1gaurav:
cmsNAMEDCOLORLIST* NamedColorList :
 692 struct _cms_NAMEDCOLORLIST_struct {
 693
 694     cmsUInt32Number nColors;
 695     cmsUInt32Number Allocated;
 696     cmsUInt32Number ColorantCount;
 697
 698     char Prefix[33];      // Prefix and suffix are defined to be 32 characters at most
 699     char Suffix[33];
In this structure Prefix & Suffix are of 33 length that means atmost 32 characters and a null termination.
In function : Type_NamedColor_Write
local variables, prefix & suffix are 32 in length and not 33, but strncpy is done for 32 elements in which last element '31' is not null.
then suffix[31] = prefix[31] = 0; is done to make 31 of each as null. this may loose 1 element.
Proper way is to make this array also as 33 instead of 32.
Another, null termination is added for Root, otherwise it will case stack buffer overflow read.


mm2:

PR #58, opened 2015-06-16
Remove tc to fix gcc error.

ya1gaurav:
Currently not using "tc" variable of algorithm by softsurfer.
GCC will report an error.
Commented tc so that it can be utilised in future if needed.


mm2:
Looks Ok, thanks!

PR #59, opened 2015-06-16
lIsLinear is being unused

ya1gaurav:
if cmsIsToneCurveLinear is false, it is written to be excluded.
Adding it in if loop to be excluded.


mm2:
lIsLinear is currently not used, and this is how it has to be. Thanks.

PR #61, opened 2015-10-06
Null Check mismatch

ya1gaurav:
Memory is allocated for NewElem ->Tab.T but NewElem ->Tab.TFloat is checked for Null.
Corrected mistake in current patch.


mm2:
Merged, thanks!

PR #63, opened 2015-10-16
Set LD_LIBRARY_PATH in testbed so just-built LCMS is tested

gnu-andrew:
When 'make check' is run, ./testcms is run in testbed/Makefile. This uses the default library path, and so will use an existing LCMS 2 installation or fail if LCMS 2 is not already installed. This patch fixes the Makefile so the tests can be run after build, but before install, to test the just-built LCMS 2.

I haven't re-generated the checked-in autotools files as they were built using a different version to the one I have locally and running them generates a raft of spurious changes.


mm2:
Thanks for the patch. I will incorporate your changes as soon as I can test it on several Linux platforms.

PR #67, opened 2016-02-27
Fix handling of Unicode descriptions for older profiles

mheinzler:
These commits allow to use Unicode descriptions with older profiles (version < 4). Before, that was not possible because they were incorrectly converted to ASCII and the Unicode representation of the description was just ignored.


mm2:
Thanks for the fix. I'm investigating how this would affect buggy profiles, which in v2.0 are legion.

ValZapod:
Please note that unicode descriptors are NOT supported in v2. 

https://color.org/security/android_bug.xalter

> Unicode data types are used to store the profile description and copyright information. These are v4 data types which are undefined in v2.

tdung (approved):

beastmode846:
nice

mm2:
It toke

PR #68, opened 2016-03-10
Fix writing of desc tag for version 2.4 profiles

beku:
The bug can be seen in Argyll's collink tool with version 2.3 profiles
from lcms. collink checks the string length inside the desc tag and compares
with the count field. collink reports then an error and stops. This patch takes
the real text length and puts it into the desc's count field, which makes
collink happy. The formal bug can be seen in Oyranos git too. Checked with
SampleICC which corrects the text count field by converting to XML and
back. Specification ICC.1:2001-04 (v2.4.0) added an example, which
suggests to do padding if any at end of the tag.

Reproduce:
$ collink lcms_profile_with_dmnd_tag_v2.icc second.icm output.icc

The change uses strlen() to obtain the count of chars.


mm2:
Thank you Kai-Uwe, nice fix!

PR #72, opened 2016-04-01
Possible double free fix.

ya1gaurav:
_cmsFree(Src ->ContextID, GammaTables); and cmsStageFree(ObtainedCurves); will be called again in case of error, so if the referred memory is already freed , it must be set to NULL, to avoid possible double free problem.


mm2:
Already fixed, thank you.

PR #73, opened 2016-04-01
Fix memory leak in cmsIT8LoadFromMem

ya1gaurav:
Before return from current function, memory pointed by it8->MemoryBlock is not freed & hence will cause memory leak.


mm2:
Thanks for your patch. Unfortunately it is not needed since cmsIT8Free already does this functionality.

PR #80, opened 2016-06-10
Warnings cleanup

mircomir:
I try to remove warnings by touching less code as possible.

gcc 5.3 and msvc 2013 warning clenup. Probabily cmsUNUSED_PARAMETER should be only cmsUNUSED (that should be created)... 
The "dereferencing type-punned" fix is not beautifull but it works. A better fix could be (not tested) a double cast by using a char\* cast.


mm2:
Old, no longer needed.

PR #83, opened 2016-06-29
UNREACHABLE_CODE.TERMINATION in vprf.c

g-coder:
Statements at line:253, line:258, line:263, line:295, line:334 are unreachable beacuse FatalError() function is called before the return statement. Program exits from FatalError() function because exit(1) is called from it.
return statement can be removed.


XhmikosR:
@mm2: any reason why such PRs aren't merged yet?

mm2:
Because different compilers and different code analyzers think differently about exit(). From strict point of view, this is a plain function call, so I could do a function called "exit()" that does something different. 
If I remove the return, other compilers will complain so I am not sure on what to do.

XhmikosR:
I wasn't aware about other compilers. If that's the case I guess it's not worth adding conditionals here.

mm2:
This is a compiler issue, the lcms code is ok.

PR #84, opened 2016-06-29
UNREACHABLE_CODE.TERMINATION in linkicc.c

g-coder:
return statement at line:291 is unreachable because Program exits from Help(0) function as exit(0) is called from it.
return Help(0) can be replaced with Help(0).


mm2:
compiler issue, the lcms2 code is ok.

PR #86, opened 2016-07-08
Added check to return value of function

g-coder:
Return value of a function 'cmsPipelineGetPtrToFirstStage' is dereferenced at cmsvirt.c:1145 without checking.
Added check to return value of cmsPipelineGetPtrToFirstStage(LUT).


mm2:
Code base is largely out of sync

PR #87, opened 2016-07-25
Document two typos in the definition of the cmsPlugin.. constants

phillipberndt:
There are two typos in the definitions of `cmsPluginTransformSig` and `cmsPluginMutexSig`. It's too late to change the constants, but the documentation can be updated safely.


mm2:
Ups, sorry, I cannot change this without breaking ABI. We have to live with it

PR #91, opened 2016-08-03
return value of function is checked

g-coder:
Added check on the return value of function 'cmsPipelineGetPtrToLastStage' at cmsopt.c:1078.


mm2:
Thanks, I've added an extra check

Issue #70, opened 2016-03-29
lcms writes uninitialized strings when writing named colors


thomas-weber:
Hi, this is Debian bug 815248, https://bugs.debian.org/815248. 
Debian is trying to move to reproducible builds (https://wiki.debian.org/ReproducibleBuilds), that is the ability to reproduce byte-identical packages.

lcms is currently using some uninitialized strings when writing named colors, leading to non-identical builds. I attach a patch that rectifies this. Please note that this is not the original patch from the above Debian bug report - the bug submitter changed some more things. Justification for these changes can be found in the bug report. I attach the original patch as well, with a "ORIG" marker in the filename.
Original patch:
[dont-write-uninitialized-memory-for-color-strings-ORIG.patch.txt](https://github.com/mm2/Little-CMS/files/193625/dont-write-uninitialized-memory-for-color-strings-ORIG.patch.txt)

Changed patch with less changes:
[dont-write-uninitialized-memory-for-color-strings.patch.txt](https://github.com/mm2/Little-CMS/files/193722/dont-write-uninitialized-memory-for-color-strings.patch.txt)

mm2:
already fixed

PR #95, opened 2016-08-31
Fix writing of arbitrary chars into ColorantTable Name

beku:
The root array gets filled by strcpy for the count of chars until
zero terminator. The bytes after the string are undefined, but gets
written into the bytestream. Memory debuggers like valgrind complain
about use of uninitialised values. As well a written profile might
contain random bytes, which have to be tracked by the profile ID.
Tools like iccToXml might by affected.

The memset initialises the "root" array, avoids the random values and
fixes the warnings.


beku:
Issue #70 might provide some more fixes

PR #102, opened 2016-10-05
Fix cmdStageAllocMatrix parameter swap

kcwu:
For cmdStageAllocMatrix, InputChans is length of Matrix, OutputChans is
length of Offsets. The original code will allocate NewElem->Offset with
length Cols=InputChans (cmslut.c:417). This results in heap buffer
overflow later.


kcwu:
I found a bug -- offset buffer allocated with wrong length. 

However, I am not sure which variables (ptr or size? caller or callee?) are
swapped. Since I don't understand what lcms is calculating, this fix is based on
my common sense best guess.

Please help review it. Thanks

PR #108, opened 2016-11-29
avoid fixed number LUT optimization on inf values

kcwu:


mm2:
Hi, I'm not taking this change because I wish to fix the infinite condition at much lower level. Will put a fix in cmsEvalToneCurveFloat()

PR #116, opened 2017-02-03
Fix windows static builds

ismail:
After the latest changes CMSEXPORT and CMSAPI were undefined for Windows static (non-DLL) builds.


mm2:
my... I accepted a pull request without checking all combinations. Thanks for fixing

PR #121, opened 2017-05-03
Ensure any open ICC profile file handles are closed on exec

lovell:
Hello, thank you for all the continued great work on the ever-useful Little CMS.

You may be aware that lcms2 is used by [libvips](https://github.com/jcupitt/libvips), which in turn is used by [sharp](https://github.com/lovell/sharp).

I've had reports that if a process using sharp and therefore lcms2 spawns a child process via exec, that process appears to inherit any lcms-opened file handles - see https://github.com/lovell/sharp/issues/674#issuecomment-289832272 for the background.

This PR ensures any file handles opened by lcms2 for reading are closed on exec using the `e` flag of `fopen`, available since glibc 2.7 (e.g. Ubuntu 8.04). I haven't been able to test this on Windows or *BSD, but I believe the former and (recent versions of) the latter should ignore this.

I've not added the `e` flag to files opened for write purposes as I suspect that a much less common use case, although I'd be happy to update this PR if you'd prefer a consistent approach.

Cheers,
Lovell

### DIFF
--- a/src/cmsio0.c
+++ b/src/cmsio0.c
@@ -385,7 +385,7 @@ cmsIOHANDLER* CMSEXPORT cmsOpenIOhandlerFromFile(cmsContext ContextID, const cha
     switch (*AccessMode) {
 
     case 'r':
-        fm = fopen(FileName, "rb");
+        fm = fopen(FileName, "rbe");
         if (fm == NULL) {
             _cmsFree(ContextID, iohandler);
              cmsSignalError(ContextID, cmsERROR_FILE, "File '%s' not found", FileName);


mm2:
This is a very interesting feature, and I am now investigating on which platforms the "e" option works. 
Thanks for the hint.

VirtualTim:
Interesting, but it looks like `e` is a Glibc extension. So it won't work on most platforms.
https://man7.org/linux/man-pages/man3/fopen.3.html#NOTES

lovell:
@VirtualTim Thanks for reviewing this. My current understanding is that the `e` flag is still planned for addition to the POSIX standard. glibc and musl both support `e`, as do the libc runtimes provided by the various *BSD since at least ~2013.

I've just checked the Windows docs and it uses `N` to specify that "the file is not inherited by child processes", so I guess we'll need to make this conditional wrapped with `#ifdef CMS_IS_WINDOWS_`.

macOS doesn't currently support this flag and appears to ignore it. When this becomes a POSIX standard then macOS will have to support it, so hopefully this is future proof.

Are there any other platforms you're aware of where either `e` or `N` are unsupported and might cause a problem?

mm2:
I must confess I totally missed this PR since... wow 2017???? OMG
Sorry, I will give it a try. What if I preserve the extra letters passed as parameter to cmsOpenProfileFromFile? Actually I only have "r" and "w" but could allow "we" and "re" for example.

mm2:
This is an Historical PR, from 2017!

So finally and by 6ae2e99a3535417ca5c95b602eb61fdd29d294d0 I'm adding support for close on exec
Now you can add "e" when calling cmsOpenProfileFromFile(), i.e 

`
handle = cmsOpenProfileFromFile("profile.icc", "re");
`

This just passes the parameter to fopen(), so its up to the caller to check whatever underlying system supports it.

lovell:
Muchas gracias Marti!

PR #124, opened 2017-06-03
Added Swift Package Manager Support

colemancda:
Small change to enable library to be built with [`swift build`](https://github.com/apple/swift).


colemancda:
@mm2 Is this something you would be interested in merging? Its a small change and makes compiling with LLVM a breeze, as well as easy Swift integration. Here is the library I built around LittleCMS https://github.com/PureSwift/LittleCMS

mm2:
Nice, but unfortunately adding support for new build systems means maintaining it and now I have no time for that. The goal is to leave lcms stable enough to be closed. Thanks for the collaboration, I would suggest to create a new package or something like that.

PR #128, opened 2017-07-17
Fix fopen() memory leak

ya1gaurav:
after fopen() , fclose() is not done so fd will remain open, hence will cause memory leak.


mm2:
I don't understand this patch. Why are you closing the file after the fopen? How the stream is supposed to be read if the file is closed? Also this would remove the write lock so an external user could modify the file while being parsed. Probably I'm missing something.

ya1gaurav:
Maybe I understand wrong:
But  FileNest is local pointer & I could not find where it is being read. I thought it is just to check if path is correct & file is present or not.

mm2:
Ah, ok, FileNest is an alias pointer for the top of stack of includes, see above:

FileNest = it8 ->FileStack[it8 -> IncludeSP + 1]

PR #129, opened 2017-07-17
Change to strncpy to silent static analyser.

ya1gaurav:
Various static analyser tools report not to use strcpy & use strncpy instead.
In this case as Name, Prefix & Suffix are coming as argument & if their size is less that src size, security issue could occur.
So, better to use strncpy.


mm2:
Looks good, I'm accepting the pull request, but modified the file in my side because you changed other unrelated parts of the source. Thanks for the patch.

ya1gaurav:
Other changes are indentation fixes.

mm2:
The code is wrong, please note sizeof(Suffix) is the size of the pointer (4), not the size of the storage array (33)

PR #131, opened 2017-07-18
Replace atoi() by safe function strtol()

ya1gaurav:
The atoi function is not thread-safe and also not async-cancel safe.

### DIFF
--- a/src/cmscgats.c
+++ b/src/cmscgats.c
@@ -1670,11 +1670,11 @@ void WriteHeader(cmsIT8* it8, SAVESTREAM* fp)
                     break;
 
             case WRITE_HEXADECIMAL:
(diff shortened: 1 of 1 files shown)


mm2:
Basically is a different use case. atoi() ignores the rest of the string after the integer number and leaves errno untouched, which is the behavior I want. The CGATS parser is not reentrant by design.

The C99 standard says this also:

7.20.1 The functions atof, atoi, atol, and atoll need not affect the value of  the integer expression errno on an error. If the value of the result  cannot be represented, the behavior is undefined.

ya1gaurav:
I recommend  to do bound check for value obtained from atoi().
It will avoid possible underflow & overflow situations.

mm2:
Thanks.

atoi() is only used on CGATS parser, which is not used by ICC profiles. 
This is a sort of helper for storing measurements on utility tools, not 
in the color management part.

PR #133, opened 2017-08-21
Constify some big arrays to allow to compiler put them in read only section.

kasper93:
Hi,

Pretty small change I had locally for a long time. Just marks some constant arrays as const letting compiler put them in r/o section. 

Regards,
Kacper


XhmikosR:
Also @mm2 I just run a Coverity scan, you should have got an email with the new defects :)

PR #142, opened 2017-11-19
src/cmsnamed.c: fix Multilocalized test on big-endian platform

trofi:
Noticed single test failure on pwerpc64, powerpc and PA-RISC
(all plaforms are big-endian):

```
Checking Multilocalized profile ...FAIL!
Multilocalized profile:
```

It's caused by endianness-depedent result of
strTo16 / strFrom16 functions.

Tweaked those functions to always construct
big-endian Country and Language codes.

Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>


mm2:
Fixed in GIT, thanks!

PR #147, opened 2017-12-13
Reindent some lines, fixing gcc indentation warnings.

sebras:
LittleCMS now compiles without warnings using:
CFLAGS=-Wall ./configure && make


mm2:
Looks great, thank you

PR #154, opened 2018-03-07
Fix some typos

stweil:


mm2:
Thanks for the code. I agree on all changes, except on endianness/endianness, where it seems there is no clear consensus, see some links on important manufacturers:

http://www.yourdictionary.com/endianess
https://www.linux-mips.org/wiki/Endianess
https://www.ibm.com/support/knowledgecenter/en/SSAL2T_9.1.0/com.ibm.cics.tx.doc/reference/r_endianess.html
https://software.intel.com/en-us/node/628919
https://stackoverflow.com/questions/9971369/little-vs-big-endianess-how-to-interpret-the-test

I would prefer to not change the spelling because there is an API function using this name.

stweil:
There is no disagreement that the only correct form is _endianness_ which is very often misspelled (even by important manufacturers). I am not a native speaker and had misspelled it myself for many years.

Of course the API cannot be simply changed. Other projects also had typos in API functions. Typically this is handled by providing a correctly spelled alias and deprecating the variant with the typo.

stweil:
I fixed https://www.linux-mips.org/wiki/Endianess now.

mm2:
You changed the wiki, and this is so fun you deserve the merge ! X-) Thank you so much.

stweil:
My last contributions on linux-mips.org were many years ago, so finding by password was the hardest part. :-) I say thank you, too, for providing this nice little library.

PR #175, opened 2018-12-07
Fix #174: Cast error in cmsxform.c reported by GCC 8

### LINKED ISSUE #174: Cast error in cmsxform.c reported by GCC 8
Little-CMS 2.9 imported in OpenJDK gives compilation error with GCC8:

cmsxform.c:839:50: error: cast between incompatible function types from ‘_cmsTransform2Fn’ {aka ‘void (*)(struct _cmstransform_struct *, const void *, void *, unsigned int,  unsigned int,  const struct <anonymous> *)’} to ‘void (*)(struct _cmstransform_struct *, const void *, void *, cmsUInt32Number,  cmsUInt32Number)’ {aka ‘void (*)(struct _cmstransform_struct *, const void *, void *, unsigned int,  unsigned int)’} [-Werror=cast-function-type]
                                    p->OldXform = (_cmsTransformFn) p->xform;
                                                  ^

In current version of the file it is
https://github.com/mm2/Little-CMS/blob/master/src/cmsxform.c#L810

Possible fix is to add intermediate (void*) cast.

dchuyko:


dchuyko:
For failed platforms I see only build host configuration errors:
configure: error: C compiler cannot create executables

PR #176, opened 2018-12-15
Fix double free of curve data structure.

sebras:
When the call to CurvesAlloc() fails in OptimizeByJoiningCurves()
the cleanup code will try to free the ObtainedCurves twice. This
is because ObtainedCurves has also been inserted into the
pipeline by the call to cmsPipelineInsertStage().

This commit NULLs the ObtainedCurves variable immediately after
the insertion into the pipeline was succesful and also moves
later use of ObtainedCurves earlier to compensate for setting
ObtainedCurves to NULL. This avoids the double free.


sebras:
It compiles nicely with `CFLAGS=-Wall ./configure && make -j` and `make check -j` runs like before.

Issue #144, opened 2017-11-23
register keyword is deprecated in C++11 and removed in C++17


VojtechFried:
When I try to compile our software that uses Little-CMS with a C++17 enabled Visual Studio I get warnings, that "register" keyword was removed from the standard. There are many ways how to deal with this warning, but the easiest for us would be that register usage is removed in lcms headers. What do you think?

mm2:
Hi, from the tutorial:
"Little CMS 2 is meant to be portable to any C99 compliant compiler."

lcms is C, not C++. It can be compiled as C++, but this is not a requirement. 
BTW, some C compilers  for embedded systems makes extensive use of "register" keyword so I cannot just remove it.

DominikDeak:
Sorry to bump the issue. However, I do feel that conditionally exposing the `register` keyword upon complier checking would be very useful. Little CMS is a great library and thus many C++ projects rely on it. The `lcms2.h` header already has tests and macros to accommodate platform specific details, such as `CMSAPI`, or `CMSEXPORT`. The `register` keyword could be given a similar treatment, as shows in the example below:

```C
#if defined( __cplusplus )
   #define CMSREGISTER
#else
   #define CMSREGISTER register
#endif

// Sampling
typedef cmsInt32Number (* cmsSAMPLER16)   (CMSREGISTER const cmsUInt16Number In[],
                                            CMSREGISTER cmsUInt16Number Out[],
                                            CMSREGISTER void * Cargo);

typedef cmsInt32Number (* cmsSAMPLERFLOAT)(CMSREGISTER const cmsFloat32Number In[],
                                            CMSREGISTER cmsFloat32Number Out[],
                                            CMSREGISTER void * Cargo);

```
I'm happy to create a pull request, if the above strategy is acceptable.

bnason-nf:
Hi @mm2,

I would also like to see the change suggested by @DominikDeak. The issue for me is that I am compiling a C++ source file using the C++ 17 standard, and it includes lcms2.h, which results in a compile failure. It's possible to work around this with a `#define register` before the include, but that's pretty ugly, and I'd prefer if this was addressed inside the lcms2 header. Can we make a pull request for this?

Thanks,
Ben

DominikDeak:
> It's possible to work around this with a #define register before the include

@bnason-nf: If you are using clang, there are three methods to temporarily work-around the problem:

[...]

PR #178, opened 2019-01-17
Provide CMSREGISTER define to allow the use of lcms2.h in C++17 code

nooploop:
This resolves the issue: https://github.com/mm2/Little-CMS/issues/144

I took the approach suggested here: https://github.com/mm2/Little-CMS/issues/144#issuecomment-409448211


mm2:
Already in codebase

PR #186, opened 2019-09-24
Buffer overflow fix.

ya1gaurav:
BuildColorantList() concatenates 32 bit data in each iteration. The max value of iteration can be 16.
Thus buffer overflow can occur with storage of 128 bytes. It would need 512 bytes in worst case.

### DIFF
--- a/src/cmsps2.c
+++ b/src/cmsps2.c
@@ -1414,7 +1414,7 @@ int WriteNamedColorCRD(cmsIOHANDLER* m, cmsHPROFILE hNamedColor, cmsUInt32Number
     cmsUInt32Number i, nColors, nColorant;
     cmsUInt32Number OutputFormat;
     char ColorName[cmsMAX_PATH];
-    char Colorant[128];
+    char Colorant[512];
     cmsNAMEDCOLORLIST* NamedColorList;
 
 


thoger:
@ya1gaurav Can you clarify under what conditions may this overflow occur?  The description mentions concatenation of 32 bit data in each iteration in the `BuildColorantList` function, but I only see it handling 16 bit data and even with maximum value further restricted.

The relevant code `BuildColorantList` is:

https://github.com/mm2/Little-CMS/blob/lcms2.9/src/cmsps2.c#L1395-L1403

```
    for (j = 0; j < nColorant; j++) {

        snprintf(Buff, 31, "%.3f", Out[j] / 65535.0);
        Buff[31] = 0;
        strcat(Colorant, Buff);
        if (j < nColorant - 1)
            strcat(Colorant, " ");

    }
```

The string generated by `snprintf` is what's added to `Colorant[]` in each iteration.  `Out[]` is defined as `cmsUInt16Number Out[]`, hence each `Out[j]` should be a 16 bit unsigned integer value, i.e. a value from the 0 ... 65535 range.  After it's divided by 65535.0, the output of `snprintf` should be strings `0.000` to `1.000` (which is consistent with the information in the comment for the `BuildColorantList` function).  `nColorant` (the number of loop iterations) is limited to `cmsMAXCHANNELS` (16).  With each formatted number being 5 characters long plus one character for space separator or trailing \0, the total space required should be (5+1)*16 = 96.  Hence the original `Colorant[]` size should have been sufficient.

Did you have any reproducer to trigger the overflow, or was it found via code inspection?

mm2:
I realized that after accepted the request. Anyway it is harmless, as this is just stack memory which is recovered on function exit.

You have good eye on finding this kind of things. I was pretty sure nobody noticed the "fix" was not needed at all.

thoger:
Marti, thank you for the quick response!

I actually noticed that the similar discussion already happened under commit 9735225.  I had not seen that before commenting here, as I came to this PR directly from the commit d1fcbdc.

Note that I started looking at this fix after it got applied to the LittleCMS code bundled in OpenJDK as part of the July security update:

http://hg.openjdk.java.net/jdk-updates/jdk11u/rev/917fdf37c538

This commit includes additional change to work around an Xcode / clang bug.  It seems that was the cause of whatever issue OpenJDK developers ran into, and not the insufficiently sized `Colorant[]` buffer.

PR #187, opened 2019-09-24
Preserve trailing null

ya1gaurav:
Write the trailing null chracter also on stream or on memory.


mm2:
it does not pass the tests, CGATS gots corrupted if this change is applied

PR #194, opened 2020-01-02
Ensure compatibility with mingw

hmaarrfk:
io.h needs to be included when using mingw compiler.


mm2:
Thanks for your contribution,
Unfortunately io.h is not standard and breaks several supported platforms. 
https://stackoverflow.com/questions/2211030/failing-to-compile-a-project-missing-io-h-file 
lcms2 library compiles fine under MinGW, although it is not yet officially supported. The only file that fails is the profile zoo testbed, which is not needed anyway.

PR #198, opened 2020-03-20
Fix ifdef that protects adding sinf/sqrtf

tsdgeos:
Doing #if (THING < 1400) is true when THING is undefined
so we first need to check if THING is defined


mm2:
No longer needed

PR #206, opened 2020-04-20
Fix tificc compile warning

kmilos:
tsize_t is signed, and can be 32 and 64 bit wide

### DIFF
--- a/utils/tificc/tificc.c
+++ b/utils/tificc/tificc.c
@@ -28,6 +28,7 @@
 #include "lcms2_plugin.h"
 #include "tiffio.h"
 #include "utils.h"
+#include <inttypes.h>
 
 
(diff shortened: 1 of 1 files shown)


kmilos:
Just realized this is incomplete, need to handle the format string as well to make truly portable

mm2:
Thanks, unfortunately this adds dependencies on a #include <inttypes.h> header and ptrdiff_t  type, that is not used anywhere else. I have commited some changes that removes the warnings and also checks for TIFF with strip/tile sizes too big.

kmilos:
No probs, thanks for reviewing

inttypes.h was only for the PRIdPTR format string, ptrdiff_t is already defined in stddef.h alongside size_t

PR #210, opened 2020-04-25
Make module definition file BCC only

kmilos:
[Not needed for MSVC as __declspec(dllexport) is already used](https://docs.microsoft.com/en-us/cpp/build/reference/module-definition-dot-def-files), and doesn't really belong in source files anyway


mm2:
Thanks, but this has to do with name decoration (mangling) .
Not sure about vs2019, but for others .def files are needed to control the exported name. Also, its a good check that the API is what is supposed to be.

kmilos:
I understand you want to document and track your API.

However, the name mangling is a non-issue for the provided MSVC projects (you use __cdecl and extern "C"), and if by any chance the user turns on CMS_USE_CPP_API, your .def file doesn't really work (I think?) since it doesn't provide the [aliases to decorated names](https://docs.microsoft.com/en-us/cpp/build/reference/exports):

> Because the MSVC compiler uses name decoration for C++ functions, you must either use the decorated name internal_name or define the exported functions by using extern "C" in the source code.

kmilos:
Apologies, I somehow assumed __cdecl was used without checking, when you in fact have __stdcall and hence indeed need the .def file for the 32-bit MSVC build. Something to consider for v3 perhaps ;)

PR #214, opened 2020-05-16
Cleanup PostScript CSA Generation

d1saster:
As reported some time ago [1], some PS interpreters have issues with the CSA code lcms2 emits
when gamma tables are encountered in the icc profile. This patch picks up the ideas from the
mentioned mailing list thread and implements them in lcms2.

Testing happened mostly with the icc profiles provided directly in the testbed folder and those generated by the testcms program. The results are all parseable by ghostscript. It should not be left unmentioned that this does **not** cover the modified code path in `EmitCIEBasedDEF`! It would be nice if somebody with an appropriate icc profile could test that one.

Generally it would be nice if we could merge this patch (or some other solution) into lcms2, since this is still a show stopper for a proper adoption of lcms2 in the PS output device of the poppler library [2]. The latter is not only central to the pdf-printing support within the Okular pdf viewer, but through the cups-filters [3] also central to linux printing as a whole.

[1] https://sourceforge.net/p/lcms/mailman/message/33182987/
[2] https://gitlab.freedesktop.org/poppler/poppler/issues/125
[3] https://github.com/OpenPrinting/cups-filters/blob/master/filter/pdftops.c#L504

### DIFF
--- a/src/cmsps2.c
+++ b/src/cmsps2.c
@@ -459,12 +459,28 @@ void EmitLab2XYZ(cmsIOHANDLER* m)
     _cmsIOPrintf(m, "]\n");
 }
 
+static
+void EmitSafeGuardBegin(cmsIOHANDLER* m, char* name)
+{
+    _cmsIOPrintf(m, "%%LCMS2: Save previous definition of %s on the operand stack\n", name);
+    _cmsIOPrintf(m, "currentdict /%s known { /%s load } { null } ifelse\n", name, name);
+}
 
+static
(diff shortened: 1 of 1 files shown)


mm2:
Fantastic! 
I did some work in this direction, but your changes does include all I was planning and even more.  
I would like to add your name in the list of contributors, it is that ok to you? What name should i put? or you prefer the nickname?

d1saster:
Wow, that was a fast merge :) You can use my real name (Philipp Knechtges), if you want, for the contributors list. But let me also take the opportunity to thank you for the incredible work you did over the years to start such a cool project and keep it going for such a long time.

PR #222, opened 2020-06-29
Update DLL resource info

kmilos:
Addresses https://sourceforge.net/p/lcms/mailman/lcms-user/thread/001401d6463d%248b576c00%24a2064400%24%40ProDigitalSoftware.com/#msg37041611

Bump file info to 2.11 and aldo reconcile Borland and VS resource files

### DIFF
--- a/Projects/VC2015/lcms2.rc
+++ b/Projects/VC2015/lcms2.rc
@@ -46,7 +46,7 @@ BEGIN
             VALUE "CompanyName", "Marti Maria"
             VALUE "FileDescription", "lcms color engine"
             VALUE "FileVersion", "2.12.0.0"
-            VALUE "InternalName", "lcms"
+            VALUE "InternalName", "lcms2"
             VALUE "LegalCopyright", "Copyright � Marti Maria 2021"
             VALUE "OriginalFilename", "lcms2.dll"
             VALUE "ProductName", "LittleCMS color engine"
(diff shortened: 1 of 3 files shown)


kmilos:
Looks like line endings changed to Windows CRLF for some files, should be ok (as expected/written by VS)

kmilos:
Undid the line ending change and rebased after 2.12 info change

mm2:
this is not needed anymore

kmilos:
Please check again, I have updated the PR after the .11 fixes...

There is still a small detail lcms->lcms2 suggested to align with Borland DLL resource info.

kmilos:
@mm2 Include this small cosmetic for the 2.13 release please? (You'll need to also update the new Projects/VC2022/lcms2.rc)

mm2:
Just for curiosity sake... why you want the dll internal name changed?
 Nobody uses that and in the file description comes "lcms color engine"

kmilos:
Not strictly necessary, but it doesn't hurt to be consistent: Borland .rc already sets it to "lcms2" so presumably there are DLLs out there that identify themselves as "lcms2". It's not a bad thing to clearly differentiate from older lcms v1.x wherever possible, not just the file name; for example, [this InternalName could be queried programmatically](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.fileversioninfo.internalname?view=net-6.0).

mm2:
Ok, this is now done.
Have a Merry Xmas!

kmilos:
Thanks, you too!

PR #224, opened 2020-07-14
Fix library version for cmsGetToneCurveParams

jmglogow:
Commit cb737b0fde11fcbb4b5cac00801f96284cf6f26d added a new
function just before the 2.11 release. But the release didn't
update the library version, so just do this to be 2.1.0.

### DIFF
--- a/configure
+++ b/configure
@@ -2406,9 +2406,9 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 #  6. If any interfaces have been removed since the last public release,
 #     then set age to 0.
 #
-LIBRARY_CURRENT=2
-LIBRARY_REVISION=10
-LIBRARY_AGE=0
+LIBRARY_CURRENT=3
+LIBRARY_REVISION=0
+LIBRARY_AGE=1
 
 
 # Obtain system type by running config.guess
--- a/configure.ac
+++ b/configure.ac
@@ -33,9 +33,9 @@ AC_CONFIG_MACRO_DIR([m4])
 #  6. If any interfaces have been removed since the last public release,
 #     then set age to 0.
 #
-LIBRARY_CURRENT=2
-LIBRARY_REVISION=10
-LIBRARY_AGE=0
+LIBRARY_CURRENT=3
+LIBRARY_REVISION=0
+LIBRARY_AGE=1
 
 AC_SUBST(LIBRARY_CURRENT)dnl
 AC_SUBST(LIBRARY_REVISION)dnl


mm2:
You are right, there was a glitch in 2.10 that forced me to release a 2.11 fews days after. Unfortunately, version number was not updated accordly. This will be naturally solved when releasing 2.12, which is scheduled in 6 months aprx. Changing now to 2.11 would only make things worst.

Otherwise stamping the version number to 3.1 in generated configure script would not help.

mm2:
I have double checked the issue you report and the version number and configure.ac was changed to 2.11 *before* the 2.11 release tag, see 17eb080324a9f16e0e7ab37bbcda7ae42b946294 
The new function was added two days after the release, so everything seems ok. The only glitch was in the lcms2.h header

jmglogow:
This doesn't change the version number to 3.1, but to 2.1. Please read the comment just before these variables in configure.ac. This patch applies the rules 3-5, because 2.11 added an interface but didn't change the ABI.

And this is important, if you want to use the new function cmsGetToneCurveParams, because you can't detect it from the versioning information otherwise. Currently v2.10 and v2.11 have the same library name, but different code, including the new function, which shouldn't happen. Commit cb737b0fde11fcbb4b5cac00801f96284cf6f26d, which introduced  cmsGetToneCurveParams, is before the 2.11 tag, otherwise I wouldn't care.

jmglogow:
To be precise, it changes the SO version from 2.0.10 to 2.1.0. You can read it all up in https://autotools.io/libtool/version.html:

> For Linux, for instance, while the last two values map directly from the command-line, the first is calculated by subtracting age from current.

mm2:
I see. Thanks for reporting. I want those fields sincronized with version number, so this function will be undocumented for a while

jmglogow:
So I checked the download, if it differs from the git repo. And it does :-( Still the function is in the header of the SF code tarball of 2.11. And the tag is after the function was added (from `git log`):

```
commit 17eb080324a9f16e0e7ab37bbcda7ae42b946294 (tag: 2.11)
Author: Marti Maria <marti.maria@littlecms.com>
Date:   Tue Jun 16 19:07:33 2020 +0200

    Documentation now in odt, results in HUGE space savings
    
    And this is source code as well

commit 2906fdde31df4fd2140470ed7a517975f64966a0
Author: mm2 <marti.maria@littlecms.com>
Date:   Mon Jun 15 14:43:20 2020 +0200

    configure for 2.11

commit cb737b0fde11fcbb4b5cac00801f96284cf6f26d
Author: Marti Maria <marti.maria@littlecms.com>
Date:   Mon Jun 15 14:37:33 2020 +0200

    Add get tone curve params function
```

I've created my tarball with `git archive -o /tmp/lcms2.tar --prefix=lcms2-2.11-git/ 2.11` and then run a diff to the SF download, resulting in:

```
$ LANG=C diff -urN lcms2-2.11{,-git}
Binary files lcms2-2.11/testbed/new.icc and lcms2-2.11-git/testbed/new.icc differ
```

No idea, if that is an additional problem, but eventually just for the test suite. Otherwise the 2.11 tarball is identical to the tag and includes the function.

Anyway, I've reported the problem. The code, which needs lcms2, doesn't use the new function, so I don't depend on it and I don't need to adapt my configure check.

Thanks for your fast replies; I appreciate this. ATB.

mm2:
Binary files lcms2-2.11/testbed/new.icc and lcms2-2.11-git/testbed/new.icc differ

This is a test profile created by testbed, with time stamp. Is only for the testbed and makes no other function than test. I think the script that creates the release also runs the testbed and modifies this file. Sorry about that.

Regarding the function, this is just a wrapper to access private data because a user asked for that. For nearly 10 years nobody has needed it, so I hope that would not be a big stopper.
 
I realise the 2.10/2.11 mixup messed out things. I hope to fix all together when 2.12, bringing everything in sync. Now changing version numbers would make things worse, a big mess and a pain for packagers, so I'm trying to minimise damage. I apologize for inconveniences.

jmglogow:
The only sensible fix would be to do a 2.12 release with the fixed SO version. And personally I would also include pull request #222 or a patch like this to fix the DLL resources. I never wanted to propose changing the 2.11 release; sorry, if you had this impression from the conversation. That would indeed create many more problems.

And just an other suggestion: please provide a sha256sum generated file with a release, even better if that was also gpg-signed. And it would be good, if the SF source tarball would be the same then the `git archive`.

mm2:
Good point,  I am adding it to wishlist. Thank you!

PR #225, opened 2020-07-15
Remove and ignore Xcode & macOS temp files

kmilos:
According to https://github.com/github/gitignore/tree/master/Global


mm2:
Thanks, this is a still work in progress... the mac folder will be deleted when Xcode project were complete. Working on the old project makes no sense anymore, it was done for powerpc in a very old machine.

kmilos:
Sure, but polluting the git (or any repo) database with unnecessary files can affect its size and speed, so it's better to avoid it as soon as possible.

mm2:
if you want to take a look, the project I have just committed is supposed to be the good one.

kmilos:
Hm, I was expecting less file changes - I guess I should've merged the other way (into my branch first), but it should still work out when merged

mm2 (approved):

PR #226, opened 2020-07-25
Handle Null Pointer in Type_ColorantTable_Read

ya1gaurav:
Fix Null return by cmsAllocNamedColorList


mm2:
Thanks, this is not necessary at all because the allocated  named color list is at most 16 entries. If you system is such stressed that malloc cannot afford 100 bytes, the segmentation fault is the minor of your worries. Anyway I have added a check for code complitude by 843cd158283d6e83588527e1020bee5ec63800b3.

PR #227, opened 2020-07-25
Handle Null pointer in PrelinOpt16alloc

ya1gaurav:
In case _cmsCalloc return NULL, avoid crash due to null pointer dereference.
Free the earlier allocated memory if any.


mm2:
Thanks for reporting. I have added a check. See 843cd158283d6e83588527e1020bee5ec63800b3

PR #230, opened 2020-09-29
Do not include autogenerated files in source control

hughsie:
Every time a user types 'make' in the source tree there are a huge number of
dirty files as the Fedora-generated ones are a tiny bit different from other
distros.

It also really messes up automated licence checks as the upstream is including
files with a very different licence to MIT.

Just include the correct magic in configure.ac so that all the files get
generated properly when running autogen.sh.

Also, actually compile the fast_float plugin rather than relying on a dummy
makefile to avoid an error.


mm2:
Hi Richard, if I understand correctly, you propose to get rid of 'configure' and other related scripts and let users to generate it by themselves by using autotools. Is tat right?
I'm afraid this will add autoconf & automake dependencies to the package. Not everybody has that. And in many places, they have   old versions. As you can see travis is broken in all builds, that means build compatibility is broken, and most of the automated build scripts will not work anymore... I understand the issue, but just forcing to have autotools installed seems a big requirement to me. 
Maybe I'm missing something... Is there any way to do what you want without breaking travis?
Thanks!

hughsie:
> you propose to get rid of 'configure' and other related scripts and let users to generate it by themselves by using autotools

Not quite. I propose removing them from git, but they're included in the tarball. So in the case the user downloads the .tar.gz they can just do `./configure && make` as before, but if they download git then they have to do `./autogen.sh && make` -- which also means that git users need aclocal and the m4 helpers installed.

> forcing to have autotools installed seems a big requirement to me

I think it might be installed already :)

Also, I did not notice the travis script; lets switch this to use `./autogen.sh && cat config.log && make && make check` and see what happens :)

mm2:
Ok, I've been thinking in this for a while. The issue, at least for me, is the amount of yet-existing scripts that uses ./configure as a way to create the Makefile. This is the basic installation, as stated by autotools itself and what many users do expect:

https://www.gnu.org/software/automake/manual/html_node/Basic-Installation.html#Basic-Installation

Alas, I keep a number of sandboxes with several linux distros/versions which I use to check before doing a new release. This is very important for some commercial packages I'm caring of. If I remove the configure script then I have to multiply every test distro by the number of possible autotools installed, which makes the sandbox collection unpractical to say the best. Also, the manuals have to be updated :(

Then, I am thinking on a different approch that hopefully would solve your issue and keep a good karma. The idea is to add to autogen.sh a toggle, something like "--clean".

In this way, packages that would like a "clean" distribution have to invoke "./autogen.sh --clean" first.
That's not a big deal, because they have to use ./autogen.sh anyway, so we will have the same old behavior by default, plus a "clean" mode for those that want to explore licenses, etc. whithout unwanted scripts. That would not break anything because there is no other change, and modifications are minimal. This is new functionality that have to be specifically enabled by using "--clean".
What do you think?

hughsie:
> yet-existing scripts that uses ./configure as a way to create the Makefile

And this would continue to work for anyone that uses the release

> If I remove the configure script then I have to multiply every test distro by the number of possible autotools installed

Really? At least on fedora running "./configure && make" overwrites the provided libtool files and means that you're testing with what Fedora provides rather than the static ones you're providing.

In a more general sense, keeping 50k LOC of untagged files in source control isn't exactly best practice...

mm2:
> And this would continue to work for anyone that uses the release

Sure, but not for those that are using the repository for latest. If I remove the scripts, I am forced to do special handling for official releases, which will be different from repository, the git tags for releases would no longer pointing to the exact release files. The release packages created by github would differ from the ones created by me.

> Really? At least on fedora running "./configure && make" overwrites the provided libtool files and means that you're testing with what Fedora provides rather than the static ones you're providing.

Ouch, I didn't try Fedora. The reason to keep configure and others like libtool is to have them tested on a number of known distributions/versions. Now you are reporting Fedora at least overwrite some of those files. Bad for Fedora, but other does not and I think keeping a script that is tested and therefore marked as "safe" is not a bad idea at all. This is the reason for those 50K that are 0.4% of lcms2-2.9, for example.

But still, you didn't tell me if the --clean option would work for you, is there anything that makes it not suitable?

hughsie:
> if the --clean option would work for you

It doesn't solve the mishmash of licences and generated files tho.

mm2:
> It doesn't solve the mishmash of licences and generated files tho.

Admitedly It is not perfect, but regenerating the files will get you exactly the same license issues 
that getting from the distribution. i.e. config.guess is GPL no matter comes from lcms package or from your 
computer. A different matter is whatever license explorers would be confused, but the few I have handled 
had the "smartness" to know that was autotools. In the case of less smart license explorers, the --clean would help. I think this is really a problem of autotools and not restricted to lcms. Maybe what I should do in a future is to change to a more modern build system, ninja or so.

hughsie:
Just look at your commit from yesterday; 02fa7a4461faf5eaa6b487bcf608720987416406 -- it was just touching two lines in two files:

    [hughsie@localhost lcms2 (master %)]$ git show 02fa7a4461faf5eaa6b487bcf608720987416406 | diffstat
     configure           |  120 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
     configure.ac        |    2 +-
     testbed/Makefile.am |    2 +-
     testbed/Makefile.in |    2 +-
     4 files changed, 121 insertions(+), 5 deletions(-)

I'm not going to tell you how you should maintain your baby (after all, you've been doing it successfully for years!) but I thought I should say something...

mm2:
Yes, of course. That was my first attempt to fix the make dist target to check if your approach could be automated somehow.
Please don't get me wrong, I see a lot of value in getting rid of all those spurious files. It is just a matter of being extremely cautious when doing so. For instance, m4 folder cannot be deleted at all because some scripts are not part of autoconf. Some are even tailored.  

I really would be happy to have something smaller and more portable, but it will take some time. I'm not merging the pull request directly, but I will progressively adding your changes. The idea would be configure will be provided only on "official" releases. Now I'm trying to figure out how to make travis happy on macs (mac OS is one of my primary goals).

Thanks for the idea!

hughsie:
>  I will progressively adding your changes.

Excellent, thanks for considering. Please don't think I'm being critical, maintaining lcms for all these years is appreciated by a huge number of people!

mm2:
That would be hopefully implemented by 5ef8c971244fe4e8e376afdcfad67899a73a9285
Still waiting travis, but seems to work.

mm2:
Tried this weekend... got lots of emails of failing automated test systems like jenkins and fuzzer. Chromium is failing, for example. Many private repositories are complaining as well. I cannot afford that.
I'm rolling back the files. In any case there is now a toggle on autogen.sh

./autogen.sh --distclean

This erases all files that autotols can regenerate. Feel free to use this script to get a simplified package, but the default will be to include those files.

PR #244, opened 2021-02-12
configure.ac: fix static linking with tiff

ffontaine:
Use pkg-config to retrieve the dependencies of tiff (e.g. xz) to fix the following static build failure:

```
libtool: link: /home/buildroot/autobuild/run/instance-2/output-1/host/bin/xtensa-linux-gcc -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -mlongcalls -mauto-litpools -Os -static -fvisibility=hidden -static -static -o tificc tificc.o ../common/xgetopt.o ../common/vprf.o  ../../src/.libs/liblcms2.a -ltiff -ljpeg -lz -lm -lpthread
/home/buildroot/autobuild/run/instance-2/output-1/host/opt/ext-toolchain/bin/../lib/gcc/xtensa-buildroot-linux-uclibc/9.3.0/../../../../xtensa-buildroot-linux-uclibc/bin/ld: /home/buildroot/autobuild/run/instance-2/output-1/host/xtensa-buildroot-linux-uclibc/sysroot/usr/lib/libtiff.a(tif_lzma.o): in function `LZMAEncode':
tif_lzma.c:(.text+0x168): undefined reference to `lzma_code'
```

Fixes:
 - http://autobuild.buildroot.org/results/07c5ca780bcdbfcd7cad6502345f1553ce17bdc3

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -330,7 +330,7 @@ then
 	AC_MSG_RESULT(no -- some components failed test)
 	have_tiff='no (failed tests)'
     else
-	LIB_TIFF='-ltiff'
+	PKG_CHECK_MODULES(PC_TIFF,libtiff-4,LIB_TIFF=$PC_TIFF_LIBS,LIB_TIFF='-ltiff')
 	LIBS="$LIB_TIFF $LIBS"
 	AC_DEFINE(HasTIFF,1,Define if you have TIFF library)
 	AC_MSG_RESULT(yes)


mm2:
Thanks for the patch, however this is what I get on my Debian:

checking for gztell in -lz... no
checking if ZLIB package is complete... no
./configure: line 13576: syntax error near unexpected token `PC_TIFF,libtiff-4,LIB_TIFF=$PC_TIFF_LIBS,LIB_TIFF='-ltiff''
./configure: line 13576: `	PKG_CHECK_MODULES(PC_TIFF,libtiff-4,LIB_TIFF=$PC_TIFF_LIBS,LIB_TIFF='-ltiff')'

Any idea of what is wrong?

mm2:
Also, I have checked the actual script and found it to work well. 

Maybe you didn't install libtiff-dev?

Try this:

sudo apt-get install libtiff-dev
./configure --disable-shared --enable-static
make
sudo make install
nm /usr/local/bin/tificc

You will see the tificc utility comes statically linked with no issues

mm2:
> in function `LZMAEncode':
>tif_lzma.c:(.text+0x168): undefined reference to `lzma_code'

This is an issue of libtiff, lcms does not call those internal libtiff functions. Chances are that the static tiff library is compiled with a special toggle to prevent including LZW compression (which was under a patent many years ago).

ffontaine:
xz is an optional dependency of libtiff which can be enabled or disabled through `--{dis,en}able-lzma`.
This dependency has been added to libtiff more than 9 years with:
https://gitlab.com/libtiff/libtiff/-/commit/d338f7fe2290da7e528f37b7df6ff7878ce1add0

If libtiff internally uses xz/lzma functions, lcms has to link with xz or the build will fail. However, as you have quickly closed this PR, I'll just disable static building of lcms on buildroot instead of applying a patch that will never be upstreamed.

mm2:
Please note lcms has NO dependencies on libtiff. Only tifficc, which is a sample, uses this library.
Then, adding pkg-config dependency for a sample is an overkill.  I would not create -or use- a package maintainer for a sample.

Disabling static building is ok to me.

PR #249, opened 2021-03-15
configure.ac: fix --without-fastfloat

ffontaine:
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -127,12 +127,15 @@ AC_ARG_WITH(fastfloat,
             AS_HELP_STRING([--with-fastfloat],
                            [build and install fast_float plugin, use only if GPL 3.0 is acceptable]),
             [
-                with_fastfloat='yes'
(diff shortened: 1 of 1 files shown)


mm2:
Thanks for the PR
Could you please explain  which functionality is added or which bugs are fixed with your modification?
Thanks

ffontaine:
Sorry but I thought it was pretty obvious: when the user provides `--without-fastfloat`, it expects that fast_float plugin will be disabled. Currently, this is not what is happening, fast_float is enabled if the user provides `--without-fastfloat` because `$withval` is never used by configure.ac. As fast_float is licensed under GPL 3.0, it seems quite important to use `--without-fastfloat` and not depends on a default value.

mm2:
Thanks! This now make sense

Issue #170, opened 2018-07-26
Data races


jan-wassenberg:
Hi, thank you for lcms!
FYI tsan (thread sanitizer) from clang reports two race conditions
in multithreaded code that opens and creates profiles:

```  Write of size 8 at 0x7f3a9ec4dc38 by thread T10:
    #0 cmsCreateProfilePlaceholder src/cmsio0.c:504 

  Previous write of size 8 at 0x7f3a9ec4dc38 by thread T7:
    #0 gmtime
    #1 cmsCreateProfilePlaceholder src/cmsio0.c:504
```

gmtime is not re-entrant because it uses static data. Would gmtime_r be an option?

The second is more straightforward:
```  Write of size 8 at 0x7fbbf621f5e0 by thread T6 (mutexes: write M2734):
    #0 cmsXYZ2xyY src/cmspcs.c:99
    #1 cmsD50_xyY src/cmswtpnt.c:42
    #2 cmsCreateXYZProfileTHR src/cmsvirt.c:572

  Previous write of size 8 at 0x7fbbf621f5e0 by thread T7:
    #0 cmsXYZ2xyY src/cmspcs.c:99
    #1 cmsD50_xyY src/cmswtpnt.c:42
    #2 cmsCreateLab2ProfileTHR src/cmsvirt.c:479
```

Again we have shared static data in
```
const cmsCIExyY* cmsD50_xyY() {
    static cmsCIExyY D50xyY;
    cmsXYZ2xyY(&D50xyY, cmsD50_XYZ());
```

As a workaround, we can just take a lock around all transform/profile creation functions, though that's a bit disappointing given their THR suffix.

DominikDeak:
> As a workaround, we can just take a lock around all transform/profile creation functions, though that's a bit disappointing given their THR suffix.

I don't think think Little CMS API calls are designed to be thread safe as a shared resource, even if you use THR suffix.

I suggest you to create a CMS context separately as a thread local resource and then work on profiles that way. Don't share them between threads.

jan-wassenberg:
Thanks for the suggestion. We do use [thread-local contexts](https://gitlab.com/wg1/jpeg-xl/-/blob/master/jxl/color_management.cc#L1447), the issue seems to be that lcms has shared mutable state (e.g. D50xyY) outside the contexts.

mm2:
[...]

PR #263, opened 2021-06-04
Ensure threadsafe alternative of gmtime is used

kleisauke:
Hello, thanks for maintaining LCMS.

As discussed in https://github.com/mm2/Little-CMS/issues/170, `gmtime` uses internal storage that is static, which means that it's not threadsafe. This PR ensures that threadsafe alternatives of this function are used (if available).

This has been successfully tested on the [JPEG XL test suite](https://github.com/libjxl/libjxl/blob/main/doc/building_and_testing.md) under ThreadSanitizer without requiring a mutex in the top-level APIs (see https://github.com/libjxl/libjxl/issues/53).

Cheers,
Kleis


mm2:
Looks great, thank you!

PR #264, opened 2021-06-09
Use a global mutex to serialize calls to gmtime

kleisauke:
After PR https://github.com/mm2/Little-CMS/pull/263, data races could still occur on systems where neither the `gmtime_r` nor `_gmtime64_s` (Windows specific) functions are available. This PR ensures that calls to `gmtime` are serialized with a global mutex in case these thread-safe functions are missing.

See https://github.com/libjxl/libjxl/pull/112 for context.

### DIFF
--- a/src/cmsio0.c
+++ b/src/cmsio0.c
@@ -485,6 +485,11 @@ cmsIOHANDLER* CMSEXPORT cmsGetProfileIOhandler(cmsHPROFILE hProfile)
 	return Icc->IOhandler;
 }
 
+#if !defined(HAVE_GMTIME_R) && !defined(HAVE__GMTIME64_S)
+// A global mutex used to serialize calls to gmtime.
+static _cmsMutex _cmsTimeMutex = CMS_MUTEX_INITIALIZER;
+#endif
+
 // Creates an empty structure holding all required parameters
 cmsHPROFILE CMSEXPORT cmsCreateProfilePlaceholder(cmsContext ContextID)
 {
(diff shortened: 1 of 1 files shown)


src/cmsio0.c
```diff
@@ -485,6 +485,11 @@ cmsIOHANDLER* CMSEXPORT cmsGetProfileIOhandler(cmsHPROFILE hProfile)
 	return Icc->IOhandler;
 }
 
+#if !defined(HAVE_GMTIME_R) && !defined(HAVE__GMTIME64_S)
+// A global mutex used to serialize calls to gmtime.
+static _cmsMutex _cmsTimeMutex = CMS_MUTEX_INITIALIZER;
```
veluca93, line 490: would it be better to have this as a local static variable?

kleisauke: Good question. According to https://stackoverflow.com/a/14107497, that would only be safe for C++11 and later versions (unless I misinterpreted that). I made it global static just to be sure.

veluca93:
Right, and I believe C is not any different here either...

mm2:
Thanks for the patch. Unfortunately this is not going to work. See cmsplugin.c lines beginning on 793 and lcms2_internal.h lines after 211. Making mutex really portable is hard. I will try to reuse the global mutex used to handle context memory management to create a safe get time function. Anyway, this is seldom used, only when the user creates an ICC profile, and this is rare.

mm2:
Thanks guys, using your idea in commit 79e8395cd405daa9fc99d7ab154762718587887a solves the issue.

kleisauke:
Great! Thanks for the follow-up; commit 79e8395cd405daa9fc99d7ab154762718587887a looks neater. I'll close.

PR #279, opened 2021-09-08
Linear RGB also comes in 32-bit and floating point

amyspark:
Hi @mm2,

In https://github.com/mm2/Little-CMS/commit/6b04ee0bab3ebe6f2c487195bac7b86efadd58d3 you added an automatic kill switch to disable optimization in case of a linear RGB space. In doing so, 32-bit and/or floating-point color spaces were left out, which broke Krita's out of gamut warnings as well as palette docker in the presence of OCIO.

This commit adds those two types of space to the check.

Fixes Krita bugs 442004 and 437429.

/cc @Lynx3d


mm2:
The issue was originated by something completly different. Thanks for reporting.

Issue #247, opened 2021-03-10
Fastfloat plugin gives odd results with cLUT carying RGB display profiles


therahedwig:
Hi,

In current Krita master, we've got the fastfloat plugin enabled, and it works great with matrix shaper profiles (like, 16bit float is now the slowest bitdepth, as 32bit float is blazing past it!), but I discovered that my display profile, which was generated with Argyll doesn't seem to work with it.

So, in Krita, we split up the image into tiles, which we then separately input into the color conversion. The image to display conversion seems to return broken or completely transparent for some of the tiles we put in, and the results are kinda arbitrary, giving me the feeling like there's a race condition happening somewhere. Once I turn off optimizations, everything works.

![image](https://user-images.githubusercontent.com/1759191/110701817-f0ae7a80-81f1-11eb-8667-97d2527c9d3f.png)

(I cannot identify why some of these images work and some don't, besides that the images that work have 1:1 or 4:3 ratio)

Another thing, which is a little bit more subtle compared to the weird results: The transparency is gone for tiles that did have transparency. Is the copyAlpha flag maybe not taken into account?

I suspect that this has something to do with the fact the display profile has luts, as simple matrix shaper profiles work just fine.

[Here's my display icc profile inside a zip file](https://github.com/mm2/Little-CMS/files/6118965/B139HAN03_0-1.2018-05-17-16-31.2_2-F-S-XYZLUT-MTX.zip)

And [here's where we create our cmsTransforms](https://invent.kde.org/graphics/krita/-/blob/master/plugins/color/lcms2engine/IccColorSpaceEngine.cpp#L43), if that helps.

mm2:
Thanks for reporting. I am giving big priority to this issue. Probably will ask you more information, keep in touch.

mm2:
Tried hard and I cannot reproduce it. 
Maybe has something to do with compiler's optimization? pointer aliasing?
Your code seems ok, and the profile is ok as well.

therahedwig:
I'll ask around during the meeting tomorrow. There's always a chance we're doing something unique with the tiles, or that there's indeed optimizations involved.

Lynx3d:
This is just a shot in the dark, but could it have to do with the rather unusual use of BGRA channel order?

It just came to my mind because I noticed the dither formats LCMS offers only include TYPE_RGBA_8_DITHER and TYPE_ABGR_8_DITHER so I was bummed I couldn't try them in krita.

[...]

PR #280, opened 2021-09-14
Fix copy alpha type mismatch in FloatCLUTEval

Lynx3d:
One more alpha channel bug that somehow slipped my testing when preparing PR #271.

A missing type cast caused a blank canvas in Krita when using float color space in combination with a CLUT display profile, so still related to issue #247.

All good things come in threes...


mm2:
Thanks for all the effort you are putting on this.

PR #284, opened 2021-10-14
Add initial Meson build system

xclaesse:
[Meson build system](https://mesonbuild.com/) is a modern and fast build system, easier to use and maintain than autotools. This PR is a really small initial support for building lcms2 with Meson.

Having a Meson build system has the advantage that lcms2 can be built as subproject of other projects that needs it, such as openjpeg. It simplify greatly building softwares on platforms that does not have proper package manager, such as Windows or when cross building.

This meson.build file I'm submitting here is being added into Meson's wrapdb https://github.com/mesonbuild/wrapdb/pull/200, but I would be more than happy to move it upstream.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,37 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  Ubuntu:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install packages
+        run: |
+          sudo apt-get -y install build-essential python3-pip ninja-build
+          pip install meson==0.59.2
+
+      - name: Build Ubuntu
+        run: |
+          meson setup build
+          meson compile -C build
+          meson test -C build
+
+  VisualStudio:
+    runs-on: windows-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install packages
+        run: |
+          pip install meson==0.59.2
+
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Choosing whether to build the DLL or the static library is selected as a meson core option: `-Ddefault_library=static` (valid options are static, shared, or both -- and it defaults to shared).

The dependencies on zlib and tiff could be made optional by switching them to options recorded in a second file `meson_options.txt`, and looking for them based on the value of the option:

```meson
zlib = dependency('zlib', required: get_option('zlib'))
if zlib.found()
    cargs += '-DHasZLIB=1'
endif
```
So if that was disabled, it would log e.g. "dependency zlib skipped: feature zlib disabled".

pkg-config is never required, but it will be attempted for dependency lookups that permit pkg-config lookup, by default all dependencies, which may be marked as optional in which case failure to find a pkg-config program will cause the dependency to be not-found and treated as missing.

You can also add dependencies using `meson wrap install zlib`, this will create a file `subprojects/zlib.wrap` and the zlib source code will be downloaded and built privately *if* it cannot be found on the system (for example pkg-config is not installed) -- this `meson wrap install` stuff is exactly why it is nice to have this PR by the way. ;) You can also force it to use the private copies with `--wrap-mode=forcefallback` in which case pkg-config is never even looked for.

xclaesse:
> On windows DLL, what about the .rc file where version and copyright are placed?

Did not know there was one, added now.

PR #295, opened 2021-12-30
Fix various typos

luzpaz:
Found via `codespell -q 3 -L ake,ba,gord,htmp,maked,pich,relativ`


mm2:
Thanks! 
Some of the fixes are on autotools, so I cannot change them and others are not quite right "build-in" vs. "built-in", so I'm changing the strings by hand on the files rather than merging the PR. Thanks again for the collaboration!

mm2:
already integrated. Thanks!

PR #296, opened 2022-01-16
Change define WIN32_LEAN_AND_MEAN for windows.h in GCC

Jamaika1:


mm2:
Thanks for your contribution. I would like, however, to underderstand the motivation of this change. Defining this symbol to 1 seems OK to me, many other software components use that. Take a look to OpenGL gl.h, for example. Windows.h accepts the current definition to 1 and it works as expected.
So, why you would  change this line?

Jamaika1:
GCC is picky program. I use static libraries. When there are two different definitions gcc starts to redefine especially combining c and c ++ files. Originally there isn't one in windows.h.

mm2:
This is on lcms2_internal.h which nobody but the lcms2 library should include. Aside the open source there are other proprietary parts using openGL and your change clashes with gl.h because the definitions are different. There is no benefit on the open source part at all.

mm2:
No further comments on that, and no need of the change, so I close the PR.

PR #305, opened 2022-02-13
simplify the CMS_IS_WINDOWS_

usta:
It was checking again and again with ifdefs instead of just adding previous defined(...) parts

### DIFF
--- a/include/lcms2.h
+++ b/include/lcms2.h
@@ -174,17 +174,10 @@ typedef cmsUInt32Number      cmsU16Fixed16Number;
 typedef int                  cmsBool;
 
 // Try to detect windows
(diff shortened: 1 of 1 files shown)


mm2:
Thanks, this is in this way for clarity sake. First line is the real detection, which is compiler agnostic, for gcc or intel compiler for instance.
Remaining blocks are for forcing on different compilers. They are below the main detection line to enforce the idea that they only works on windows. Functionally is same, is just a matter of code style, in fact the #ifdef _MSC_VER is not needed at all.

mm2:
Thanks to you. I look forward to see a collaboration from you in a future. Best regards.

PR #306, opened 2022-02-17
Fix trivial typos in testbed

luzpaz:


mm2:
Good finding. Thank you!

PR #321, opened 2022-05-01
cmsgamma: Use float instead of double for a significant speedup.

sliedes:
When profiling RawTherapee, I started to look into lcms2 performance. One very simple thing that can be done is to use `powf()` (the float version of `pow()`) in `DefaultEvalParametricFn`. This already gives a nice speedup. All tests still pass, and I assume this should still be enough precision for color transforms in general.

Some of the functions there were double precision. To avoid unnecessary conversions between floats and doubles, I changed some of them to float.

There are probably other places in the code base where it would be safe to change to use float versions of math functions (also fabsf, logf, ...); this only changes the cmsgamma module, as that was the bottleneck I identified.

Significant-looking changes in performance tests:

```
32 bits on Matrix-Shaper profile           : 6.43 -> 7.75 MPixel/sec
32 bits on SAME Matrix-Shaper profiles     : 5.14 -> 7.18 MPixel/sec
32 bits on Matrix-Shaper profiles (AbsCol) : 6.10 -> 7.62 MPixel/sec
32 bits on curves                          : 5.23 -> 7.48 MPixel/sec
32 bits on gray-to gray                    : 10.6 -> 12.5 MPixel/sec
32 bits on SAME gray-to-gray               : 10.6 -> 12.6 MPixel/sec
```

### DIFF
--- a/src/cmsgamma.c
+++ b/src/cmsgamma.c
@@ -310,26 +310,26 @@ Error:
 
 
 // Generates a sigmoidal function with desired steepness.
-cmsINLINE double sigmoid_base(double k, double t)
+cmsINLINE float sigmoid_base(float k, float t)
 {
-    return (1.0 / (1.0 + exp(-k * t))) - 0.5;
+    return (1.0 / (1.0 + expf(-k * t))) - 0.5;
 }
 
-cmsINLINE double inverted_sigmoid_base(double k, double t)
+cmsINLINE float inverted_sigmoid_base(float k, float t)
 {
-    return -log((1.0 / (t + 0.5)) - 1.0) / k;
+    return -logf((1.0 / (t + 0.5)) - 1.0) / k;
 }
 
-cmsINLINE double sigmoid_factory(double k, double t)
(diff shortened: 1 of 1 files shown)


mm2:
Please use the fast flat plugin if you want a speedup. Reducing precision of parametric curves would affect other things  aside color transforms. There is an API on that so it can be (and certainly are because I have several) clients that may want full precision. Anyway the speedup you obtain doing that is little when compared with the plugin.

mm2:
After further investigation, I found this breaks parametric curve plug-ins, see cmsParametricCurveEvaluator in lcms_plugin.h After this modification, existing parametric curve plug-ins no longer works.

This technically would need major version bump. Is a huge price to pay for a small speedup that is much surpassed by the fast float plug-in.

tdung (approved):

tdung (commented):
Ý

tdung (commented):
Ý

tdung (commented):
Ý

tdung (commented):
Ý

tdung (commented):
Ý

Issue #303, opened 2022-02-10
2.13.1: build time warnings


kloczek:
I'm usingg gcc 12.0.1.
Summary stats:
```console
[tkloczko@devel-g2v SPECS]$ rpmbuild -ba lcms2.spec --quiet 2>&1 | grep -- -W | sed 's/.*\[//; s/\]//' | sort | uniq -c | sort -nr
      6 -Wstringop-overflow=
      3 -Wunused-variable
      3 -Wunused-function
      3 -Wstringop-overread
      2 -Wstrict-aliasing
      1 -Wunused-result
      1 -Wunused-but-set-variable
      1 -Wmissing-braces
      1 -Wmisleading-indentation
```
And extracted warnings from stderr
```console
cmsplugin.c: In function '_cmsReadFloat32Number':
cmsplugin.c:181:15: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  181 |         *n = *(cmsFloat32Number*)(void*)&tmp;
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cmsplugin.c: In function '_cmsWriteFloat32Number':
cmsplugin.c:312:12: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
  312 |     tmp = *(cmsUInt32Number*) (void*) &n;
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
jpgicc.c: In function 'TransformImage':
jpgicc.c:945:13: warning: 'cmsSetAlarmCodes' reading 32 bytes from a region of size 8 [-Wstringop-overread]
  945 |             cmsSetAlarmCodes(Alarm);
      |             ^~~~~~~~~~~~~~~~~~~~~~~
jpgicc.c:945:13: note: referencing argument 1 of type 'const cmsUInt16Number[16]'
In file included from ../../utils/common/utils.h:41,
                 from jpgicc.c:28:
../../include/lcms2.h:1789:35: note: in a call to function 'cmsSetAlarmCodes'
 1789 | CMSAPI void             CMSEXPORT cmsSetAlarmCodes(const cmsUInt16Number NewAlarm[cmsMAXCHANNELS]);
      |                                   ^~~~~~~~~~~~~~~~
In function 'TransformImage',
    inlined from 'main' at jpgicc.c:1268:5:
[...]

PR #323, opened 2022-07-01
Fix build errors with -Wstrict-aliasing

rndxelement:
Building with the -Wstrict-aliasing C-flag resulted in build errors
in src/cmsplugin.c.
Using memcpy resolves this, as it is a correct way for
type punning in C.

Part of a fix for issue #303


Flowdalic:
You may want to consider putting the a reference to the downstream Gentoo bug, i.e. https://bugs.gentoo.org/855020, in the commit message.

mm2:
Many tanks. I sincerely appreciate your collaboration. Unfortunately I will not merge your PR for following reasons:
 
-  line 185 `memcpy(&n, &tmp, sizeof tmp);` is wrong, n already is a pointer.
- static_assert() is not part of C99.
- as its names hints, cmsFloat32Number and cmsUInt32Number have both 32 bits, if this is not the case the testbed detects it (see checkbasetypes function in testcms2.c) So the whole assert is not needed at all, because if sizes are not same whole library fails earlier.
- memcpy may not be implemented as intrinsic on some compilers, like those for PICs, therefore adding a stack frame and breaking atomicity may change behavior on threaded environments. Testing that is not trivial. 
-And more importantly, **there is nothing wrong on original lines**, the punned pointer is temporary and should be discarded by gcc.  I suspect this is a compiler bug, and as a proof of that I have fixed the warning message by  introducing a temporary pointer, that is, adding a true aliasing. This oddly seems to make gcc happy. Visual studio was not issuing any warning at all.

In general I'm pretty reluctant to add code to make compilers happy, specially when the offending code seems OK, but in this case was easy and hopefully harmless. See the changes on: 

https://github.com/mm2/Little-CMS/commit/c3d7f491e2daebda2413fb3d2935c51df1c50ac7#diff-b6710fb02050a438b53b59a1a4a0e52b9a887b3b20b4188dc4f274df7cb73750

SoapGentoo:
@mm2 whether the punned pointer is temporary or not is immaterial. You cannot dereference a pointer to a type not compatible with the original object, temporary or not.

As a compromise solution, I would suggest going with a write-read through a `union`, with is generally easier to optimize, even for embedded and PIC compilers (and implementation-defined behaviour in C). Adding more type punning to try and trick GCC into accepting this code will just cause issues down the road.

SoapGentoo:
Superseded by #324

mm2:
Superseded by https://github.com/mm2/Little-CMS/pull/324

PR #324, opened 2022-07-05
Perform type punning via union without undefined behavior

SoapGentoo:
* The previous code from c3d7f491e2daebda2413fb3d2935c51df1c50ac7
  still contains undefined behavior, since it just creates
  temporary pointer variables.


mm2:
Looks great. Thanks, please allow me a couple of days to check on several computers.

PR #330, opened 2022-08-11
Fix case when AllocBigBlock returns NULL in AllocChunk

anfedotoff:


anfedotoff:
This changes are made by this [commit](https://github.com/mm2/Little-CMS/commit/a6ef58ec3716bad934d0cacc063ef1a419f4668f).

PR #336, opened 2022-08-31
Update meson to build utils

### LINKED ISSUE #335: meson build missing option to build utils
Hi @mm2 it would be good to have the option with meson to build utils. 

Something like:

`meson_options.txt`
```
option(
  'utils',
  type : 'boolean',
  value : false,
  description : 'Enable building utils')
```

heitbaum:
### Draft
- fixes #335 

Building with `-Dutils=true` will now build the ICC utils (same as `configure; make`)

e.g.
```
Installing utils/jpgicc/jpegicc to …/install_pkg/lcms2-2.13.1/usr/bin
Stripping target 'utils/jpgicc/jpegicc'.
```

Signed-off-by: Rudi Heitbaum <rudi@heitbaum.com>

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -6,6 +6,6 @@ option(
 
 option(
   'jpeg',
-  type : 'boolean',
-  value : true,
+  type : 'feature',
+  value : 'enabled',
   description : 'Enable building with jpeg')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,0 +1,11 @@
+option(
+  'utils',
+  type : 'boolean',
+  value : false,
+  description : 'Enable building utils')
+
+option(
+  'jpeg',
+  type : 'boolean',
+  value : true,
+  description : 'Enable building with jpeg')
--- a/utils/jpgicc/meson.build
+++ b/utils/jpgicc/meson.build
@@ -1,0 +1,24 @@
+#project('jpegicc', 'c')
+
+jpegicc_srcs = [
+  'jpgicc.c',
+  'iccjpeg.c',
+  '../common/xgetopt.c',
+  '../common/vprf.c',
+  '../common/utils.h',
+]
+
+inc_common_dirs = include_directories('../common')
+
+libjpeg_dep = dependency('libjpeg', required: get_option('jpeg'))
+if libjpeg_dep.found()
+  executable('jpegicc',
+    jpegicc_srcs,
+    c_args: cargs,
+    dependencies : [liblcms2_dep, libjpeg_dep],
+    include_directories : inc_common_dirs,
+    install : true)
+endif
+
+#man_MANS = jpgicc.1
+#EXTRA_DIST = iccjpeg.h $(man_MANS)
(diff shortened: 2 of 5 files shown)


mm2:
Seems jpgicc misses  libm

utils/jpgicc/meson.build
```diff
@@ -0,0 +1,24 @@
...
+  'iccjpeg.c',
+  '../common/xgetopt.c',
+  '../common/vprf.c',
+  '../common/utils.h',
+]
+
+inc_common_dirs = include_directories('../common')
+
+libjpeg_dep = dependency('libjpeg', required: get_option('jpeg'))
+if libjpeg_dep.found()
```
eli-schwartz, line 14: This will be found even if the jpeg option is false. I suggest instead...

meson_options.txt
```diff
@@ -0,0 +1,11 @@
...
+  'utils',
+  type : 'boolean',
+  value : false,
+  description : 'Enable building utils')
+
+option(
+  'jpeg',
+  type : 'boolean',
+  value : true,
+  description : 'Enable building with jpeg')
```
eli-schwartz, line 11:
```suggestion
option(
  'jpeg',
  type : 'feature',
  value : 'enabled',
  description : 'Enable building with jpeg')
```

See the documentation for feature options: https://mesonbuild.com/Build-options.html#features

utils/jpgicc/meson.build
```diff
@@ -0,0 +1,24 @@
...
+  executable('jpegicc',
+    jpegicc_srcs,
+    c_args: cargs,
+    dependencies : [liblcms2_dep, libjpeg_dep],
+    include_directories : inc_common_dirs,
+    install : true)
+endif
+
+#man_MANS = jpgicc.1
```
eli-schwartz, line 23: The Meson way to do this is with: https://mesonbuild.com/Reference-manual_functions.html#install_man

```suggestion
install_man('jpgicc.1')
```

utils/jpgicc/meson.build
```diff
@@ -0,0 +1,24 @@
+#project('jpegicc', 'c')
+
+jpegicc_srcs = [
+  'jpgicc.c',
+  'iccjpeg.c',
+  '../common/xgetopt.c',
+  '../common/vprf.c',
+  '../common/utils.h',
```
eli-schwartz, line 8: You don't need to add headers, but it doesn't hurt. They mostly just affect generated IDE information.

The common/ source files, in general, should probably be compiled once and reused for all the utils (although this draft PR only implements jpgicc for now, you will probably want the rest eventually?).

What you could do is add `utils/common/meson.build` which:
- builds a [`static_library()`](https://mesonbuild.com/Reference-manual_functions.html#static_library),
- makes a [`declare_dependency()`](https://mesonbuild.com/Reference-manual_functions.html#declare_dependency) that links to the static library and adds the common directory as an `include_directories:`

and then link that dependency, let's call it `common_dep`, to each utility executable in the `dependencies:` kwarg.

heitbaum: got it - that's a good way to do it. I'll put that together and squash our commits. (then I'll rinse and repeat for the other icc directories.)

mm2:
This keeps failing because libm

mm2:
Any news on that? It still doesn't work

PR #340, opened 2022-10-04
meson: cleanups

neheb:
Ran through muon. Fix soversion on Windows to match def file.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,64 +1,66 @@
-project('Little-CMS', 'c',
+project(
+  'Little-CMS',
+  'c',
   version: '2.14',
   meson_version: '>=0.48.0',
 )
 
 library_version = '2.0.13'
 
+add_project_arguments('-D_POSIX_C_SOURCE', language: 'c')
+
 cc = meson.get_compiler('c')
 
-lcms2_srcs = [
+lcms2_srcs = files(
   'src/cmsalpha.c',
-  'src/cmscgats.c',
-  'src/cmserr.c',
-  'src/cmsgmt.c',
-  'src/cmsintrp.c',
-  'src/cmsio1.c',
-  'src/cmsmd5.c',
-  'src/cmsnamed.c',
-  'src/cmspack.c',
-  'src/cmsplugin.c',
-  'src/cmssamp.c',
-  'src/cmstypes.c',
-  'src/cmswtpnt.c',
   'src/cmscam02.c',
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -67,19 +69,23 @@ if host_machine.system() == 'windows'
...
+  'lcms2',
+  lcms2_srcs,
+  include_directories: inc_dirs,
   gnu_symbol_visibility: 'hidden',
-  dependencies: deps,
+  dependencies: [m_dep, thread_dep],
   c_args: cargs,
+  name_prefix: host_machine.system() == 'windows' ? '' : 'lib',
   version: library_version,
+  soversion: host_machine.system() == 'windows' ? '' : '2',
```
kmilos, line 81: Is it possible to make this MSVC specific somehow? Isn't this automatically figured out depending on platform like it is for CMake?

With current MinGW build using Autotools the default result is `liblcms2-2.dll`, i.e. it has both the prefix and the soversion. It would be good to keep this behaviour so all the dependant binaries don't have to be rebuilt.

neheb: It is but the def file is specific about the name. AFAIK if it mismatches, linking won't work properly.

kmilos: The def file is completely optional, there is no need for it when you have the import library stub (at least when we talk about MSVC and MinGW).

neheb: Sure

kmilos: FYI, if I remove those two lines, the Meson build creates identical artifacts as [Autotools for MinGW](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-lcms2?repo=ucrt64) (minus the tools and the static lib):

```
Installing liblcms2-2.dll to C:/msys64/opt/lcms2/bin
Installing liblcms2.dll.a to C:/msys64/opt/lcms2/lib
Installing C:/msys64/home/kmilos/Little-CMS/include/lcms2.h to C:/msys64/opt/lcms2/include/
Installing C:/msys64/home/kmilos/Little-CMS/include/lcms2_plugin.h to C:/msys64/opt/lcms2/include/
Installing C:/msys64/home/kmilos/Little-CMS/build/meson-private/lcms2.pc to C:/msys64/opt/lcms2/lib/pkgconfig
```

No .def file in either case.

mm2:
superseded by meson build system

PR #341, opened 2022-10-20
fast float plugin: Fix proper accessing of LCMS2 internals

amyspark:
Hi @mm2!

This PR fixes a slight mishap in the fast float plugin internals. Both LCMS and the fast float plugin share common code in its header file, which causes definitions to shadow each other when the tests are built with `LCMS_FAST_EXTENSIONS`. This is not allowed in MSVC.

### DIFF
--- a/plugins/fast_float/src/fast_float_internal.h
+++ b/plugins/fast_float/src/fast_float_internal.h
@@ -22,6 +22,7 @@
 #ifndef _FAST_INTERNAL_H
 #define _FAST_INTERNAL_H
 
+#include "lcms2_internal.h"
 #include "lcms2_fast_float.h"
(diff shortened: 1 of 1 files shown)


mm2:
Hello,

Thanks for your contribution. Unfortunately this PR breaks the rule than plug-ins should not access lcms2 core internals in any way. This is the point of having a plug-in API and plug-in documentation.  Think that plug-ins may be compiled into a DLL by using a given version of lcms2, and then used by another different version of lcms2.

Anyway you have uncovered an issue, which is the existence of  LCMS_FAST_EXTENSIONS in the lcms2 testbed. That is a remains of development and should not be there. I am removing it for 2.14

Thanks for let me know!

Issue #339, opened 2022-09-28
autoconf update required to support clang 15 on Linux (detect HAVE_FUNC_ATTRIBUTE_VISIBILITY)


emmenlau:
With lcms2.13, I get a build error when using clang 15.0.2 on Ubuntu Linux. I'm using the `autotools` based build, to create a shared library. I could trace the problem to the configure log `config.log`, where it (wrongly) reports:
```
config.log:1932:ax_cv_have_func_attribute_visibility=no
```
This leads to problems in symbol visibility further down the build!

After updating `autotools` (just using `autoreconf -fiv`), the build works fine!

Could the autotools-scripts in master be updated? I'm using `automake 1.16.5`, `aclocal 1.16.5`, `autoconf 2.71` and the like from Ubuntu 22.04, but maybe even slightly older versions are sufficient, I did not test.

kmilos:
We already had `autoreconf -fiv` for ages, but still missing symbols w/ Clang 15.0.3 on MSYS2 for the recent 2.14 release: https://github.com/msys2/MINGW-packages/actions/runs/3372986986/jobs/5597048630#step:11:666

thesamesam:
https://github.com/mm2/Little-CMS/pull/342 should fix it, in particular https://github.com/autoconf-archive/autoconf-archive/commit/083716b47c8a95dda5a58362ccd3e04021c743a5.

thesamesam:
By the way, if it doesn't fix it (although it should, and I checked with Clang), please share the full config.log + configure output here.

kmilos:
@mm2 FYI, the unrelated MSYS2 MinGW problem was solved by [this patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-lcms2/0001-fix-symbol-export.patch) (both GCC and Clang define `__GNUC__` on Windows and understand the visibility attribute)

mm2:
Thanks, I'm adding the patch in c8ed7427b83df73f3039ef570c48c6b995c2a7ab
MinGW was never officially supported. Hopefully this patch changes that.

PR #342, opened 2022-11-03
m4: sync autoconf-archive (ax_*) macros with upstream

thesamesam:
- Update the various ax_* macros (which originate from autoconf-archive)
- Switch from acx_pthread.m4 -> ax_pthread.m4

Fixes: https://github.com/mm2/Little-CMS/issues/339
Signed-off-by: Sam James <sam@gentoo.org>


thesamesam:
I don't think there's any harm in the additional stuff this readds given that it only affects tests you actually check the #defines for anyway.

PR #347, opened 2022-12-03
Assorted Meson nitpicks

amyspark:
Hi @mm2 @neheb @xclaesse,

This PR proposes several nitpicks to make the Meson toolchain feature compatible with the Autotools one.

I've **not** made attempts at being name or SOVERSION compatible, but I've tried to make as much of the version generation match the original (and thus not need to manually replace every time there's a release bump). I also completed the pkgconfig module to match the Autotools output, and completed the dependency detection to make sure it'll pick dependencies from both Meson wraps as well as pkg-config.

Fixes #336 
Supersedes #340


mm2:
A great contribution. Thank you so much.  Please allow me a day to check everything.

meson.build
```diff
@@ -31,10 +31,24 @@ if host_machine.endian() == 'big'
 endif
 
 # Check for threadsafe variants of gmtime
-if cc.has_function('gmtime_r', prefix: '#include <time.h>')
+if cc.has_header_symbol('time.h', 'gmtime_r')
```
neheb, line 34: So on MinGW in order for this to succeed, _POSIX_C_SOURCE must be passed.

neheb:
The minimum meson version is quite high. It should probably be compatible with 0.57.0 at a minimum.

mm2 (approved):
I have done some very small modifications, otherwise everything is OK. I will now run the extended testbed

mm2:
github test fails about minimum meson version. Could you somehow relax this?

plugins/threaded/include/lcms2_threaded.h
```diff
@@ -24,7 +24,7 @@
 
 #include "lcms2_plugin.h"
 
-#if !defined(CMS_IS_WINDOWS_) && defined(CMS_NO_PTHREADS)
+#ifdef CMS_NO_PTHREADS
```
amyspark, line 27: Reinstating `CMS_NO_PTHREADS` means that this plugin is effectively impossible to compile in Windows under MSVC.

https://github.com/mm2/Little-CMS/blob/c8ed7427b83df73f3039ef570c48c6b995c2a7ab/include/lcms2.h#L257-L263

`HasTHREADS` is only defined if pthreads was found.

meson.build
```diff
@@ -1,86 +1,177 @@
-project('Little-CMS', 'c',
+project(
+  'Little-CMS',
+  'c',
   version: '2.14',
-  meson_version: '>=0.48.0',
+  meson_version: '>=0.60.0',
```
amyspark, line 5: I need 0.60 because of the TIFF and JPEG lookup, as neither have a sane name with which to look them up.

For instance, libtiff is `libtiff-4` if asked through pkg-config, `libtiff` if it's a Meson wrap, or `TIFF` if it's CMake.

And libjpeg is the same, `jpeg` if it's the original, `libjpeg` if it's libjpeg-turbo and asked through its wrap or pkg-config. (This one has the added complexity of verifying the API version.)

plugins/threaded/include/lcms2_threaded.h
amyspark: @mm2 if we keep my original change, the plugin builds and passes the test suite without issue under Windows. Was blocking MSVC compatibility your intent?

meson.build
neheb: Sounds like a bug in the wrap. It should match the pkgconfig name.

neheb: ping @eli-schwartz

plugins/threaded/include/lcms2_threaded.h
mm2: Yep. That's for compiling the CMM with no thread support. Is refers NOT the phread library, but the capability of working with threads. This plugin needs threads so hence the check. Not sure if meson assumes this is the pthread library.

mm2: Forgot to comment... this is a configuration toggle defined in lcms2.h
If HasTHREADS is defined then it is forced to 1, but otherwise is up to user to decide if the CMM should be thread aware or not.

meson.build
amyspark: > Sounds like a bug in the wrap. It should match the pkgconfig name.

If built through CMake (the official way) it generates TiffTargets.cmake and libtiff-4.pc.

eli-schwartz: > For instance, libtiff is `libtiff-4` if asked through pkg-config, `libtiff` if it's a Meson wrap, or `TIFF` if it's CMake.

@amyspark for the record, this is a serious WrapDB policy violation if true -- it's flat-out forbidden.

The WrapDB's entire *reason* for existing, is to provide lookup-name parity with the official upstream project.

> Sounds like a bug in the wrap. It should match the pkgconfig name.

@neheb fortunately, it's not a WrapDB bug. The WrapDB specifies a `[provide]` section already that defines `libtiff-4 = libtiff4_dep`, and it is this that Meson uses to resolve the wrap.

`dependency('libtiff')` will fail, because it will not find any subproject providing such a dependency. As a last-ditch heuristic, it will attempt to configure `subproject('libtiff')` and check if it overrides `libtiff`, fail (because the libtiff wrap doesn't override any such thing) and, as a result of that failure, discard the libtiff subproject as unusable.

@amyspark at a minimum, you MUST remove `libtiff` from the lookup of tiff_dep, it will not do anything. The `tiff` one won't either, I'm pretty sure, because it needs to be uppercased to match CMake. In fact, libtiff itself doesn't install a cmake find_package configuration at all. (It does provide package exports, but not find_package anything. Darnedest thing I ever did see.) There's a totally unofficial FindTIFF.cmake shipped as part of the cmake sources.

I am curious if you actually tested this on a system without pkg-config but with cmake? Which makes me wonder why this project needs to support some totally unofficial ad-hoc cmake.org thing when you were happy with the version that didn't work. Maybe that means pkg-config is good enough?

I reiterate: upstream libtiff doesn't provide cmake find_package support anyway. All they provide is:
- pkg-config: `dependency('libtiff-4', method: 'pkg-config')`
- something which claims to be useful inside cmake's FetchContent exclusively

If you want libtiff upstream to provide cmake find_package support, you'll need to open a gitlab MR asking for it -- if you do so, please implement it via the name `find_package(libtiff-4)` to match the pkg-config file. :D

eli-schwartz: > And libjpeg is the same, jpeg if it's the original, libjpeg if it's libjpeg-turbo and asked through its wrap or pkg-config. (This one has the added complexity of verifying the API version.)

I'm interested in this claim too -- I just downloaded the IJG edition of libjpeg, and checked -- they provide the pkg-config name `libjpeg.pc`, just like libjpeg-turbo does.

Funny story: IJG only started providing a pkg-config file since 2018, two years after libjpeg-turbo started doing so.

Furthermore, libjpeg-turbo provides a cmake configuration only for `dependency('libjpeg-turbo', method: 'cmake')`, and IJG provides nothing at all.

Totally unofficial cmake.org stuff provides `FindJPEG.cmake`, which again needs capitalization that you did NOT add and perhaps did not test? and isn't the official way to look it up anyway.

I don't think you *can* look for libjpeg-turbo at all though, since that exposes a different ABI -- the `tj*` symbols instead of the `jpeg_*` ones. You don't want to find that at all via pkg-config, and you want to specify `modules: ` to cmake to get the correct library, so you'd have to first check for the generic pkg-config name, and then afterward check for -turbo via cmake only.

This could be handled, more or less, by adding a libjpeg dependency class to Meson and having it handle getting the cmake name correct. Projects like lcms would benefit automatically simply by optionally choosing to build them with a newer version of Meson.

meson.build
```diff
@@ -1,86 +1,177 @@
...
+      dependencies: jpeg_dep,
+    )
+    and cc.compiles(
+      '''
+    #include <stdio.h>
+    #include <stdlib.h>
+    #include <jpeglib.h>
+    #if JPEG_LIB_VERSION < 62
+    #error IJG JPEG library must be version 6b or newer!
+    #endif
```
eli-schwartz, line 95: This is... a LOT of tests just to check the version.

You will never find an IJG JPEG that old anyway, since they don't provide pkg-config until 9c. If this PR was updated to be compatible with CMake's totally unofficial FindJPEG.cmake, then that extracts the version from jpeglib.h or jconfig.h's `JPEG_LIB_VERSION` and sets it to the cmake variable JPEG_VERSION, which means that Meson will see that too.

No matter how you look at it, `jpeg_dep.version()` will provide a suitable version.

...

But also are there really people who still use the IJG version? Wow.

meson.build
```diff
@@ -1,86 +1,177 @@
...
-  link_with : liblcms2_lib,
-  include_directories : inc_dirs
+if tiff_dep.found()
+  if (
+    cc.check_header('tiff.h', dependencies: tiff_dep)
+    and cc.check_header('tiffio.h', dependencies: tiff_dep)
+    and cc.check_header('tiffconf.h', dependencies: tiff_dep)
+    and cc.has_function('TIFFOpen', dependencies: tiff_dep)
+    and cc.has_function('TIFFClientOpen', dependencies: tiff_dep)
+    and cc.has_function('TIFFIsByteSwapped', dependencies: tiff_dep)
```
eli-schwartz, line 124: Do you really need to check that all these functions exist? Are they likely to not exist? How would one get a libtiff without these?

Note that in configure.ac these checks are done only because AC_CHECK_LIB cannot tell if the tiff software is officially installed, and just does the equivalent of Meson's `cc.find_library()` -- so it is important to check that the library in question is the one you meant to find.

meson.build
```diff
@@ -1,86 +1,177 @@
...
+deps = [m_dep, threads_dep]
+
+subdir('include')
+subdir('src')
+subdir('testbed')
+subdir('utils')
+
+if get_option('samples').enabled()
+  subdir('utils/samples')
+endif
```
eli-schwartz, line 168: This is a tristate option, being misused to have two options mean the same thing. When configured with `-Dsamples=auto`, it is equivalent to disabling it entirely, when the semantic meaning is to check if required system dependencies and interfaces are available and it "can" be built.

plugins/threaded/include/lcms2_threaded.h
amyspark: > Not sure if meson assumes this is the pthread library.

It's the Autotools script that assumes so:

https://github.com/mm2/Little-CMS/blob/486b5f537b2cbb7848590f2d153afccbffee4170/configure.ac#L240-L250

The author of the Meson script assumed the (p)threads library is always available:

https://github.com/mm2/Little-CMS/blob/486b5f537b2cbb7848590f2d153afccbffee4170/meson.build#L48-L51

meson.build
amyspark: > I don't think you can look for libjpeg-turbo at all though, since that exposes a different ABI

I'm *not* looking for libjpeg-turbo, just for the libjpeg library it provides. However, said pkg-config file does *not* provide the correct (API) version in its file, hence the additional checks needed, which I ported from the Autotools script:

```
❯ cat /usr/local/lib/pkgconfig/libjpeg.pc
prefix=/usr/local/Cellar/jpeg-turbo/2.1.4
exec_prefix=/usr/local/Cellar/jpeg-turbo/2.1.4
libdir=/usr/local/Cellar/jpeg-turbo/2.1.4/lib
includedir=/usr/local/Cellar/jpeg-turbo/2.1.4/include

Name: libjpeg
Description: A SIMD-accelerated JPEG codec that provides the libjpeg API
Version: 2.1.4
Libs: -L${libdir} -ljpeg
Cflags: -I${includedir}
```

eli-schwartz: > However, said pkg-config file does not provide the correct (API) version in its file, hence the additional checks needed, which I ported from the Autotools script:

libjpeg-turbo was created by forking libjpeg 6b, so all versions of libjpeg-turbo are good enough. I suspect the checks ported from the autotools script are intended to detect IJG libjpeg 5 -- because 6b was released the same year LittleCMS was created.

Since that was back in 1998 I'm not sure how important it is to detect and properly error out in such a situation. :)

meson.build
```diff
@@ -1,86 +1,119 @@
...
+    cargs += '-DHAVE_GMTIME_S=1'
+  endif
 endif
 
-if cc.has_function_attribute('visibility:hidden')
-  cargs +='-DHAVE_FUNC_ATTRIBUTE_VISIBILITY=1'
+jpeg_dep = dependency('libjpeg', required: get_option('jpeg'))
+
+if jpeg_dep.found()
+  cargs += ['-DHasJPEG=1']
```
eli-schwartz, line 68: I grepped the sources and AFAICT it's an inefficiency of the configure.ac / Makefile.am that `HasJPEG`, `HasTIFF`, or `HasZLIB` are ever defined as compiler macros.

#348

mm2:
Almost. Still a math lib missing on ubuntu.

mm2:
Not sure about *.rc.in files.  Resource files are for windows only, and windows configuration is not handled by autoconf. Changing the build system to support configure on windows is out of scope. So, this will never be used. Or maybe is meson who creates the rc file? I confess I don't know much about meson. On the other hand, the plug-ins are intended to be linked statically, working as DLL has not been tested.

mm2:
Now there are only two issues blocking the merge:

-  An "undefined reference to symbol 'floor@@GLIBC_2.2.5'" in ubuntu-meson
-  The *.rc.in files, which are of no use but for Cygwin/Msys2 and those are unsupported.

@amyspark Could you please take a look? If you could solve those, I will be glad to do the merge. Thanks!

meson.build
amyspark: Alright, removed.

amyspark:
> The *.rc.in files, which are of no use but for Cygwin/Msys2 and those are unsupported.

Meson will use these files for all supporting Windows compilers, among them MSVC, which it is supported through your Projects folder. But there's more to it, see below.

> Or maybe is meson who creates the rc file?

That is correct. I made them as template files is to allow you to set the project version once, in the Meson file, and then the build system will update all the relevant version definitions by itself. (I've left the lcms2.h line out, which is the last instance of a hardcoded version, to ensure I didn't muck with the Autotools toolchain.)

> On the other hand, the plug-ins are intended to be linked statically, working as DLL has not been tested.

We ship the fast float plugin at Krita as a shared library for license reasons, it's been working great without any issues.

amyspark:
> Almost. Still a math lib missing on ubuntu.

Apologies, I hadn't noticed you used `-lm` there. Fixed.

mm2 (approved):
All seems ok to me.

amyspark:
> Tried to fix libm issue by myself and broke everything, sorry...

No problem, it was a simple fix: Meson takes dependencies as either a single one, or a list of them between brackets.

amyspark:
Oh, wow, I didn't know this repo had Github CI. I enabled it on my fork and have checked all build jobs pass.

https://github.com/amyspark/Little-CMS/actions/runs/3632086254

PR #348, opened 2022-12-04
configure.ac: do not define preprocessor symbols for Makefile conditions

eli-schwartz:
The dependency checks for jpeg/zlib/tiff are only used for determining which utils to add in the Makefile using AM_CONDITIONAL. There's no reason to also pass the Makefile conditional variables as C defines.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -302,7 +302,6 @@ fi
     else
 	LIB_JPEG='-ljpeg'
 	LIBS="$LIB_JPEG $LIBS"
-	AC_DEFINE(HasJPEG,1,Define if you have JPEG library)
 	AC_MSG_RESULT(yes)
         have_jpeg='yes'
     fi
@@ -343,7 +342,6 @@ then
     else
       LIB_ZLIB='-lz'
       LIBS="$LIB_ZLIB $LIBS"
-      AC_DEFINE(HasZLIB,1,Define if you have zlib compression library)
       AC_MSG_RESULT(yes)
       have_zlib='yes'
     fi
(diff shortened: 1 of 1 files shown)


mm2:
Is there any special reason to remove that? 
It is being used by extended testbed that is not on the open-source side.
I would like to know if the fact of defining those symbols affects any other package. Should not.

eli-schwartz:
The main reason is because it is confusing, see for example the Meson PR. It also extends the length of the command line by about 30 characters because the code uses DEFS instead of autoheader, so I figured it was nicer if it's not used.

I have no idea how the extended testbed works. :D Is it hooked into the autoconf code? Maybe it would be easy to simply transfer the defines to the testbed instead?

mm2:
>I have no idea how the extended testbed works. :D 
This is a detailed check I use in commercial applications. Basically, is a replacement of testcms2.c in the testbed directory. Since it checks and compares images, needs to know if TIFF/JPEG packages are installed. It uses this symbol at compile time.

> The main reason is because it is confusing, see for example the Meson PR. 
You should just ignore this. It is not used in normal operation.

> It also extends the length of the command line by about 30 characters
On Linux/Mac this is not a problem and configure.ac is only supported on Linux/Mac.

eli-schwartz:
Whether intentional/supported or not, it currently does build via configure.ac on MSYS2 for Windows. :D

mm2:
>Whether intentional/supported or not, it currently does build via configure.ac on MSYS2 for Windows. :D

Fortunate chance.  Began to work in c8ed7427b83df73f3039ef570c48c6b995c2a7ab and is not tested. Low priority since we have wsl and visual studio is free for personal use.

mm2:
Unfortunately, I cannot pick this one because it breaks the extended testbed and gives no benefits.

PR #357, opened 2023-01-24
Ensure plugins can be built and installed with Meson

lovell:
Hola Marti, this small PR makes the dependency on the `m` library explicit when building via Meson to avoid undefined reference errors during compilation/linking.

It also allows the development headers to be installed, matching the behaviour of the autotools-based build.

Finally it enables the "fastfloat" option in the Meson-based CI job so it matches the behaviour of the existing autotools-based CI job. (I'm unsure if there's a desire to also enable the "threaded" option, but this PR at least makes things consistent.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,7 @@ jobs:
 
       - name: Build Ubuntu
         run: |
-          meson setup build
+          meson setup build -Dfastfloat=true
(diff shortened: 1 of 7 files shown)


mm2:
Looks great,
Thank you!

Suer90 (commented):
- [ ] - []()

PR #358, opened 2023-01-25
Meson: skip utils with missing optional dependencies

lovell:
- `jpgicc` depends on optional `libjpeg`
- `tificc` depends on optional `libtiff-4`


mm2:
Thanks Lovell, meson build scripts are still in experimental state so any help is greatly appreciated

PR #360, opened 2023-03-02
meson: fix big endian cargs

nekopsykose:
otherwise they are a string and appending with + [""] later fails in
plugins/fast_float/testbed/meson.build:5:0: ERROR: The `+` operator of str does not accept objects of type list (['-DPROFILES_DIR="/home/buildozer/aports/main/lcms2/src/lcms2-2.15/plugins/test_profiles/"'])


hughsie:
This also fixes the Fedora s390x build, thanks @nekopsykose

PR #361, opened 2023-03-02
meson: Install psicc man page

Biswa96:


amyspark (commented):
Apologies for the mishap, it looks good!

PR #362, opened 2023-03-02
Fix compiling the samples using meson

hughsie:


amyspark (commented):
Apologies for the mishap, it looks good!

PR #363, opened 2023-03-03
Fix building on case sensitive mingw platforms

kleisauke:
Mingw headers are all lowercase, and can be used for cross compilation from case sensitive file systems.

The official Windows SDK headers aren't self-consistent wrt upper/lower case, so those headers can't be used on case sensitive systems without a layer providing case insensitivity anyway.

This matches other includes of windows.h throughout the codebase.

### DIFF
--- a/plugins/fast_float/src/lcms2_fast_float.rc.in
+++ b/plugins/fast_float/src/lcms2_fast_float.rc.in
@@ -19,7 +19,7 @@
 //
 //---------------------------------------------------------------------------------
 
-#include <Windows.h>
+#include <windows.h>
 
 1 VERSIONINFO
 FILEVERSION @LCMS2_VERSION_MAJOR@, @LCMS2_VERSION_MINOR@, @LCMS2_VERSION_MICRO@, 0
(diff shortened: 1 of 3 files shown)


amyspark (commented):
Looks good to me, but what filesystems are case-sensitive on Windows? NTFS is case-sensitive but will redirect lowercase from/to uppercase.

mm2:
Look good. Thank you!

kleisauke:
> but what filesystems are case-sensitive on Windows?

AFAIK, all file systems (FAT, NTFS, etc.) on Windows are case insensitive. So, this was only a issue when cross compiling from a case sensitive file system (and when using Meson, I wrongly assumed that these files were also used in Autotools).

mm2:
>what filesystems are case-sensitive on Windows? 
Cygwin and MSYS2 mixes tools and FS. Not very used anymore because WSL supersedes them, but some people are still tied to that because legacy systems, etc. Alas, this is only on MESON builds, using configure works fine on 2.15

PR #364, opened 2023-03-04
Roundtrip CMM, platform, and creator tags

amyspark:
This fixes generated MD5 checksums of existing profiles not matching between Windows and Unix-based systems, by letting existing profiles preserve their CMM, creator, and platform tag values.

See https://invent.kde.org/graphics/krita/-/commit/1191295a4618a93893987497e3c54e6f0c2fd025#note_634123


mm2:
Seems good, and I like the fact you get some extra information from the profile. 
Please let me check against some projects, because there is a behavior's change that may affect 3rd parties.
Thanks!

PR #365, opened 2023-03-04
Fix MSYS2 compatibility

amyspark:
Hey @mm2,

This PR groups together the fixes needed to enable CI for all Intel MSYS2 flavors:

- used concurrency groups to supersede builds quickly when force-pushing
- fixed the missing exported APIs issue I reported in the mialing list
- fixed your usage of `timespec_get` in the threaded plugin test bench, which on Windows, is UCRT-only and C11 or higher

As regards the missing private APIs-- given that you intend to use GCC's visibility to export the functions in Windows, do you want me to remove the `vs_module_defs` in that case?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,10 @@ name: Build
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.job }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
   Ubuntu:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,3 +101,56 @@ jobs:
           meson setup build
           meson compile -C build
           meson test -C build
+
+  msys-meson:
+    name: 'msys-meson ${{ matrix.sys.abi }} ${{ matrix.library }}'
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    strategy:
+      matrix:
+        sys:
+          - { abi: mingw64, env: x86_64,       compiler: gcc }
+          - { abi: ucrt64,  env: ucrt-x86_64,  compiler: gcc }
+          - { abi: clang64, env: clang-x86_64, compiler: clang }
+        library: ['shared', 'static']
+      fail-fast: false
+    steps:
+      - name: Use MinGW from MSYS
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{matrix.sys.abi}}
+          update: true
+          path-type: inherit
+          install: >-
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -97,3 +101,56 @@ jobs:
...
+      - name: Use MinGW from MSYS
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{matrix.sys.abi}}
+          update: true
+          path-type: minimal
+          install: >-
+            mingw-w64-${{matrix.sys.env}}-toolchain
+            mingw-w64-${{matrix.sys.env}}-meson
+            mingw-w64-${{matrix.sys.env}}-ninja
```
kmilos, line 131: Just use the ["pacboy" `foo:p` way of installing packages for the matching environment,](https://github.com/msys2/setup-msys2#pacboy) there is no real need to carry around this `env` variable...

Also, the `toolchain` metapackage will pull in _a lot_ of unnecessary stuff and waste CI resources (both time and watts), please use the minimal set of packages needed - [for lcms2 this are just `cc` and `pkgconf` (w/ `meson` and `ninja` of course).](https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-lcms2?repo=ucrt64)

.github/workflows/build.yml
```diff
@@ -97,3 +101,56 @@ jobs:
...
+        env:
+          CC: ${{ matrix.sys.compiler }}
+      - name: Setup compiler
+        if: startsWith(matrix.sys.abi, 'clang')
+        run: |
+          CXX=${CC/#clang/clang++}
+          echo "CC=$CC" >> $GITHUB_ENV
+          echo "CXX=$CXX" >> $GITHUB_ENV
+        env:
+          CC: ${{ matrix.sys.compiler }}
```
kmilos, line 148: Similarly, there is no need to mess w/ MSYS2 defaults, these are already set up accordingly, so the `compiler` variable can also be removed.

.github/workflows/build.yml
dragonmux: In this context, `pacboy` syntax will not work as the Action directly invokes pacman. `toolchain` here is used for reduced brittleness because you don't just need a compiler and pkg-config, you also need a pile of other dev tools for a correct environment such as various headers that are not installed by default. Which ones depends on the sub-environment, and `toolchain` guarantees you pull the correct things in.

kmilos: I don't think so, I have been using both pacboy and minimal packages needed for build in several CI actions for various projects already. Please try both of these out.

amyspark:
@dragonmux @kmilos Seems that newer setup-msys2 takes care of both of these, so I've moved forward with the requested changes.

mm2:
@amyspark  ok, I was done exactly that, you sent the PR before I could commit.

PR #367, opened 2023-03-06
Simplify MSYS2 CI jobs further

kmilos:
Current: ucrt64 static job takes 10m 17s and

```
  Total Download Size:   154.66 MiB
  Total Installed Size:  977.88 MiB
```

Proposed is 9m 27s and no additional download

```
  Total Installed Size:  653.59 MiB
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,8 +102,8 @@ jobs:
           meson compile -C build
           meson test -C build
 
-  msys-meson:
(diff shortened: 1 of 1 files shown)


kmilos:
Also, I don't think the CI failure is specific to this PR, as it already (intermittently?) happens on master already: https://github.com/mm2/Little-CMS/actions/runs/4386006557/jobs/7679480015

PR #379, opened 2023-05-04
Simplify fixed-point to double conversion routines

nico:
No behavior change.

I checked that old and new code has the same results: https://godbolt.org/z/Eo3GT7rzP


mm2:
It is a nice idea, and it seems to work on all cases I've tried. As soon as I could check a big endian machine I will accept the PR

PR #389, opened 2023-06-08
Fix PostScript output for linear tone curve

goossenp:
For profiles with a linear tone curve, the Emit1Gamma function doesn't produce any output, so the PostScript code generated by LCMS may reference names (such as "lcms2gammaproc") that it does not actually define.

This patch makes it generate an empty procedure in such cases instead.


goossenp:
Thank you for merging this small fix.

mm2:
Thanks for providing the fix. That comes from a PR many time ago that added named references because some ps interpreters runs faster in that way. Obviously that was not well tested, and since very few people uses ps nowadays the bug has been there for a while. It toke me 3 weeks to locate a real ps printer to check your fix, it works perfectly, thank you again!

Issue #391, opened 2023-06-28
Ship separate pkgconf files for plugins


kmilos:
As things stand, both `lcms2_fast_float` and `lcms2_threaded` end up in `extra_libraries` for generation of core `lcms2.pc`.

As a consequence, projects not wanting to link to these for licensing reasons cannot decouple them. Separate plugin .pc files (that then require lcms2) are needed.

mm2:
I have tried configure and  configure --with-fast-float --with-threaded 
It does the right thing. Exports the libraries that you specify in the lcms2.pc pkgconfig. 
Or this happens on the meson build?
I begin to think adapting meson was not a good idea at all.

kmilos:
> Exports the libraries that you specify in the lcms2.pc pkgconfig.

It is the same w/ Meson.

I'm saying that's not the way to go regardless of the build system - they should have standalone .pc files.

The core `lcms2.pc` should not link to them implicitly. Project wishing to use GPL plugins should use `lcms2_fast_float.pc` and `lcms2_threaded.pc` explicitly.

mm2:
Sorry, I don't get the point. If you install system wide libraries that means those libraries are intended for that system. That includes licensing. 

If you wish to develop a commercial product and therefore avoid linking with GPL'd parts, then you should not use system libraries that are out of your control. In general you should never use system libraries for any development you wish to distribute. In any case, if you don't call the initialization function, which is the only function plugins exposes, your final binary is not going to contain any of the plug-in code, either static or as shared object. 

Making independent pc files makes the product unnecessarily complex.  My desire is to promote open source and GPL licensing, and providing separate pc files would penalty open source projects and benefit proprietary ones. My goal is exactly the opposite.

So, my recommendation for anyone trying to avoid GPL3 parts is to install a local copy. It is not a big deal and  that will also prevent to break the project if the system library us updated by someone else and the new version is incompatible or buggy.

mm2:
See

PR #392, opened 2023-06-28
meson: Create separate pkgconfig files for plugins

Biswa96:
Fixes https://github.com/mm2/Little-CMS/issues/391

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -124,5 +124,16 @@ subdir('plugins')
 pkg = import('pkgconfig')
 pkg.generate(liblcms2_lib,
   description: 'LCMS Color Management Library',
-  libraries: extra_libraries
 )
+if get_option('fastfloat')
+  pkg.generate(liblcms2_fast_float,
+    description: 'LCMS fast float plugin',
+    requires: liblcms2_lib
+  )
+endif
+if get_option('threaded')
+  pkg.generate(liblcms2_threaded,
+    description: 'LCMS multi threaded plugin',
+    requires: liblcms2_lib
+  )
+endif


mm2:
Hi, please see the discussion here : https://github.com/mm2/Little-CMS/issues/391#issuecomment-1611351494

This is for me to understand why you need this. In the case you want to commercially distribute your program using the system libraries, which is something you should avoid, just not calling the plug-in init function in your program prevents any GPL'd code to be linked.  With separated pc files, you are forcing open-source clients to include a different configuration for each plug-in, which breaks compatibility with current distribution and offers nothing in exchange. Or maybe I'm missing something important?

Biswa96:
Also, the lcms2.pc has `-llcms2 -llcms2_fast_float` in Libs field. If the program is not using the plugins, `-llcms2_fast_float` is not required.

Biswa96:
Thanks for the explanation. I agree that there are indeed some negative side effects to consider.

mm2:
Thanks for checking the issue!

eli-schwartz:
> I was thinking of general use cases. For example, one can check if fastfloat or threaded plugin is available through pkgconfig files. Some projects like cairo, harfbuzz etc. have different pkgconfig files for different libraries.

I would suggest instead adding a variable `features=fastfloat,threaded` to the .pc file (in meson, pass `variables: ['features=...']` to the pkgconfig.generate() call).

People seeking to confirm which plugins are available would do the same lcms2 lookup they used to do, but this time with `--variable features` to get a list of features available.

This is also done by e.g. libcurl.pc via `supported_protocols` and `supported_features`, so it's not a novel approach.

And it has no effect on people's current usage of the pkg-config files.

mm2:
Sure, we can add a variable on pc files, and probably that would be handy for someone, but I still like the "keep as simple as possible" approach, that's "little" CMS, remember? I guess just checking for plug-in header existence would be enough, then let the liker to do its job.

kmilos:
> And what is being linked

As @eli-schwartz pointed out to me, the `--as-needed` linker option that you rely on here (and that was introduced as default not that long ago) might not be the default (or even available?) on all platforms and toolchains.

Perhaps you can then add it to the .pc file linker options if you're reliant on it (and the linker supports it) and want to "keep as simple as possible"?

mm2:
I was thinking this option relates to weak references. So that means without this flag, if I link a small program by using -lm I got all math library included, no matter I only used a single floor()? 
Anyway I still think for commercial program redistribution nobody should use system libraries but local copies

eli-schwartz:
If you statically link then the linker optimizes out anything that it can. Since you passed the unused static libraries on the command line, it does more work -- opening the archive and searching for symbols in it -- but then realizes it needs nothing, tosses it all out (or just keeps floor.o and tosses everything *else* out) and moves on.

If you dynamically link then the linker just adds a tag for the dynamic loader to load that library at runtime, which is actually needed if you use floor() -- yes, the whole library, because dynamic linking. The interesting question is what happens if you don't use any functions at all. By default the linker adds that dependency anyway, and the --as-needed flag tells it to omit the NEEDED tag in this case.

Again, this only matters when doing dynamic linking.

PR #399, opened 2023-09-06
tweak: nullify freed tags inside freeOneTag(), where it is more suitable

GerHobbelt:
nullify freed tags inside freeOneTag(), where it is more suitable than where it was done before (now we hit *both* spots where this call is used; slightly cleaner code this way)

Easiest to see what's been changed by looking at the diffs, I suppose.

---

Extracted these edits from [my own fork](https://github.com/GerHobbelt/thirdparty-lcms2.git) (based off another git repo, related to mupdf) during a manual source tree review with your master branch. Trivial stuff, really. 😊


mm2:
Sorry I cannot get this one. freeOneTag() is called twice. In one of this calls setting the pointer to null has no effect.
So, your change trades an assignment after the function call by an assignment not needed at all.

PR #400, opened 2023-09-06
Complete check for WIN32/WIN64 defines for various compilers to detect 64-bit Windows target builds (MSVC / PVS Studio)

GerHobbelt:
Ran into this one back when I was still using MSVC2019 and building 64-bit static libs; the V comment is to shut up [PVS Studio](https://pvs-studio.com/en/pvs-studio/).

Easiest to see what's been changed by looking at the diffs, I suppose.

---

Extracted these edits from [my own fork](https://github.com/GerHobbelt/thirdparty-lcms2.git) (based off another git repo, related to mupdf) during a manual source tree review with your master branch. Trivial stuff, really. 😊

### DIFF
--- a/include/lcms2.h
+++ b/include/lcms2.h
@@ -174,7 +174,7 @@ typedef cmsUInt32Number      cmsU16Fixed16Number;
 typedef int                  cmsBool;
 
 // Try to detect windows
-#if defined (_WIN32) || defined(_WIN64) || defined(WIN32) || defined(_WIN32_)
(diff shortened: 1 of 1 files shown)


mm2:
Sorry on that, but cannot get this one. You basically are adding a case for the symbol WIN64, and this symbol is not defined nor documented in windows:

https://pvs-studio.com/en/blog/posts/k0040/
https://msdn.microsoft.com/en-us/library/b0084kay.aspx
https://groups.google.com/g/microsoft.public.vc.mfc/c/UstZOsTT14U

I think adding check to a symbol that no one uses makes no sense and may even be dangerous for any project that uses WIN64 internally.

PR #403, opened 2023-09-22
Do not use the "register" keyword by default.

jpakkane:
It causes compilation errors in C++17 and newer, where the use of "register" is forbidden.

GCC and MSVC seem to be fine with it, but Xcode 15's Clang hard errors on it.

### DIFF
--- a/include/lcms2.h
+++ b/include/lcms2.h
@@ -61,8 +61,8 @@
 // Uncomment this for special windows mutex initialization (see lcms2_internal.h)
 // #define CMS_RELY_ON_WINDOWS_STATIC_MUTEX_INIT
 
-// Uncomment this to remove the "register" storage class
-// #define CMS_NO_REGISTER_KEYWORD 1
+// Comment this to use the deprecated "register" storage class
+#define CMS_NO_REGISTER_KEYWORD 1
 
 // ********** End of configuration toggles ******************************
 


mm2:
Thank you for your collaboration.
 
I am, however, not going to take this PR. Reason is, Little CMS is a library written and oriented to C99. Its primary goal is to provide a C99 API for clients. As a matter of compatibility, the library can be compiled by any C++20 compiler, but this requires some adjusts, like defining (probably by command line) some symbols. Since the used language for code and API is C99, it makes sense that the defaults for such symbols are adjusted to match C99. There are ways to overcome the warning when using a C++17 compiler. In gcc, you can use -std=c99  to enforce strict compliance. I found also that the compiler accepts "register" if the header is placed on a system folder.

On the other hand, changing this default would have catastrophic consequences for many embedded systems using C language. Some compilers, Intel for example,  and some ARMs too, implements the use of register for parameters passing. This provides a throughput boost which is required in some critical parts.  Silently moving the default to remove the register attribute would not only break ABI compatibility but also create bottlenecks on IoT devices like camaras that are using the library for YCC to Lab or multispectral handling. 

Hope that makes sense to you.

jpakkane:
> There are ways to overcome the warning when using a C++17 compiler. In gcc, you can use -std=c99 to enforce strict compliance.

This issue is not about compiling the library itself, it is about compiling a C++23 program that uses the library via `#include <lcms2.h>`, specifically using a new enough Clang. The error message says that according to the C++ standard, `register´ is a forbidden keyword.

If `register` was only used inside lcms, this would not be an issue. The fact that it is used in the public header is.

The only way to use lcms2 on such compilers currently is to manually edit the header file, which is not great.

If this can not be changed, then it would be nice if lcms2 would define that macro to be empty on platforms where `register` does not affect the ABI. At the very least there should be configure option so people could turn it off if they so chose (though obviously that is not great user experience).

> Some compilers, Intel for example, and some ARMs too, implements the use of register for parameters passing.

I created [a test on Compiler explorer}(https://gcc.godbolt.org/z/n6jdcn7nz) that has the same function but with one instance using `register` and the other not. On all ARM compilers I tested the generated assembly is identical for both of these except on compilers that refuse to compile the code at all because `register` is prohibited. Obviously this is not exhaustive and I can't speak for Intel as it's not available on CE.

Thanks,

mm2:
Texas instruments dspic compilers also uses that, but that is not the point. Even in the case no single C++ compiler would use the register keyword,  this is still a valid syntax for c99, no matter c++17 has deprecated it. The library uses c99 and provides a way to accomodate (some) other languages, but still it is a c99 thing. In your case, the option I use is 

#define CMS_NO_REGISTER_KEYWORD
#include <lcms2.h>

This is harmless o  gcc/clang and has to be double checked on embedded.

mm2:
@jpakkane Please see 42309a0cd4dde6ae7fd650c2e8513fff5b46a5fa which could solve your issue

jpakkane:
Looks good. :+1:

PR #421, opened 2023-12-17
Projects/VC2019: Convert CRLF to LF

th1722:
These files should use LF like other source files.


mm2:
Those are not source files but project files used by visual studio, no matter its internal format is XML 
I tried Visual Studio 2019 and it accept flawless the projects. So, there is no issue here to fix.

PR #426, opened 2023-12-22
Fix various typos

luzpaz:
Found via `codespell -q3 -S "*.pdf,./m4,./ltmain.sh" -L ake,clen,htmp,maked,mis,ntemp,pich`


mm2:
Great, thanks, will review in detail after Christmas.

PR #443, opened 2024-04-12
Optimize cmsOpenIOhandlerFromMem() in 'w' mode

x2018:
Let the IOhandler successfully created by cmsOpenIOhandlerFromMem always has a valid Block.


mm2:
Thanks for your contribución. Just to clarify the issue you are solving... how you do to get "Block" member  set to NULL?

x2018:
Thank you for your prompt reply.

From the view of lcms, it is meaningless to set Block to NULL.
But for the users in practice, they may tranfer NULL due to unintentional reasons such as the memroy allocation problems.

After the changes, at least lcms will be more robust within its scope.

mm2:
Sorry, I didn't explain myself. I mean, which sequence of lcms functions you call to get "Block" set to NULL? 
I am asking because in  the case it is not possible to set this value to NULL, then this check is useless, I.e, why not check if somebody has set it to 42 instead?

x2018:
I'm sorry that I misunderstood your meaning.

You are right, the first changed line of MemoryWrite() will indeed be theoretically useless, because the `Block` should not be NULL for a valid IOhandler after the second changed line takes effect.

Prior to this, as cmsOpenIOhandlerFromMem() itself is an exported function, the `Block` can be NULL if the `Buffer` passed to cmsOpenIOhandlerFromMem() is NULL in 'w' mode, since cmsOpenIOhandlerFromMem() only check it when it is 'r' mode without checking it in 'w' mode.

Because the test case crashed at memmove() in MemoryWrite(), I just made that change at the beginning.
Now I have removed it.

PR #449, opened 2024-05-02
Convert line endings from CRLF to LF ...

ben2s:
for consistency with other sources and to avoid compiler warnings.


mm2:
Already done in current sources, anyway thanks for reporting

PR #450, opened 2024-05-28
Remove redundant preproc definition

MajorMurphy:
Removed second definition of TYPE_RGBA_HALF_FLT


mm2:
Already removed in current sources, anyway thanks for reporting

PR #464, opened 2024-11-07
Disable warning

MonkeybreadSoftware:
We got a warning while building:

 warning C4244: "=": Konvertierung von "double" in "cmsFloat32Number", möglicher Datenverlust

so I added the cast for cmsFloat32Number.


mm2 (approved):
neat!

PR #466, opened 2025-01-12
fix bug detected at /Wall warning level in MSVC: comparison with the wrong variable: the slides[] pointer instead of the `nslices` count.

GerHobbelt:
:smile: as it says on the tin: 

> fix bug detected at /Wall warning level in MSVC: error C7664: '>': ordered comparison of pointer and integer zero ('_cmsWorkSlice []' and 'int')


mm2:
Congratulations! you have found a genuine bug! 
Thank you very much for the hunting and for the PR

PR #474, opened 2025-02-17
Update config.sub to 2024-01-01 version

freakboy3742:
The current main branch is using the 2018-02-22 version of config.sub from autotools. 

This PR replaces config.sub with [the 2024-01-01 version](https://git.savannah.gnu.org/cgit/autoconf.git/tree/build-aux/config.sub). 

This doesn't change any behavior for most common platforms. However, it adds support for a number of new platforms, including Apple's iOS and tvOS.

### DIFF
--- a/config.sub
+++ b/config.sub
@@ -1,12 +1,14 @@
 #! /bin/sh
 # Configuration validation subroutine script.
-#   Copyright 1992-2018 Free Software Foundation, Inc.
+#   Copyright 1992-2024 Free Software Foundation, Inc.
 
-timestamp='2018-02-22'
+# shellcheck disable=SC2006,SC2268 # see below for rationale
+
+timestamp='2024-01-01'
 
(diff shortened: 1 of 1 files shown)


mm2:
Thanks. I prefer to keep in sync with autotools/automake, despite some files are largely out of date. 
For lcms, this is only used as a way to install component so grabbing and maintaining separate parts of autoconf seems an overkill to me.

freakboy3742:
Totally understood that you may want to update all the autotools files at once; and even then, you may want to manage that change yourself rather than audit that the diff is actually what it says it is. 

I've raised this PR mostly to draw attention to the fact that the current config.sub is 6 years old, and actually imposes an impediment to development at present, as it prevents compiling the library for iOS, tvOS, and a number of other platforms.

mm2:
I have validated it across several platforms and found it works great. Thank you for the update!

PR #481, opened 2025-03-04
Fix windows thread deps detection

### LINKED ISSUE #477: When building for Windows, pthreads should not be required.
https://github.com/mm2/Little-CMS/blob/04ace9c100fc6850f09223dd6c7ad8fa634acd06/plugins/threaded/include/lcms2_threaded.h#L27-L29

LCMS uses the Win32 API directly instead of pthread, so why the hard error here? If I remove this check, I can still build LCMS withthread plugin without pthread.

Andarwinux:
Using host_machine.system() instead of is_visual_studio
fixes #477

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -82,7 +82,7 @@ if (
 endif
 
 
-if is_visual_studio    
+if host_machine.system() == 'windows'
     m_dep = []
     threads_dep = []
     deps = []


mm2:
That would break cygwin and msys2 with no other benefit. Only visual studio is currently supported for Win32 API direct access.

mm2:
> Since you're just holding your ground and refusing to admit your mistake, I'm going to assume that lcms is broken and drop it from my build.

:-D

mm2:
FYI I have tested cygwin with the configure script and it works fine with proper handling of pthread library. 
The supported scripts work ok on supported platforms.

PR #501, opened 2025-07-20
const up some static arrays

caolanm:
which enables moving them out of the .data section

### DIFF
--- a/src/cmsalpha.c
+++ b/src/cmsalpha.c
@@ -377,7 +377,7 @@ int FormatterPos(cmsUInt32Number frm)
 static
 cmsFormatterAlphaFn _cmsGetFormatterAlpha(cmsContext id, cmsUInt32Number in, cmsUInt32Number out)
 {
-static cmsFormatterAlphaFn FormattersAlpha[6][6] = {
+static const cmsFormatterAlphaFn FormattersAlpha[6][6] = {
 
        /* from 8 */  { copy8,       from8to16,   from8to16SE,   from8toHLF,   from8toFLT,    from8toDBL    },
        /* from 16*/  { from16to8,   copy16,      from16to16,    from16toHLF,  from16toFLT,   from16toDBL   },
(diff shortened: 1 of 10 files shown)


mm2:
Thanks for your collaboration, I am cherry-picking some of your changes.  I am explaining here the cases I discard:

`static const char* PredefinedSampleID[]  vs. static const char* const PredefinedSampleID[]
`
I am not using that. All arrays in all code are placed in ds. String constants are not. Changing everything to this model would require many changes and extensive testing.

`cmsIntentsList "next" pointer set to const`
This has a subtle side effect that is the linked list cannot be increased/decreased by  code anymore  because you cannot write the "next" member.  There are ways to overcome this, but at the cost of making the code a bit more complex. Not worth of the small benefit, as this is only an internal structure. Applies for other linked lists.

`cmsTagTypeHandler* TypeHandler  vs. const cmsTagTypeHandler* TypeHandler `
Type handlers are not designed to be stateless and therefore needs to be writeable. Some plugins are using this feature.

`WhitesAreEqual(cmsUInt32Number n, cmsUInt16Number White1[], cmsUInt16Number White2[] ) vs. static cmsBool WhitesAreEqual(cmsUInt32Number n, const cmsUInt16Number White1[], const cmsUInt16Number White2[] )`
I agree C is basically broken on arrays passed by value, but unless you know that arrays decay to a pointer, this syntax may confuse you. Again, I prefer clarity. 
 
`cmsUInt16Number *WhitePointIn, *WhitePointOut; vs. const cmsUInt16Number *WhitePointIn, *WhitePointOut;
`
I guess that is a typo, second pointer is not constant.

`struct _cmsOptimizationCollection_st *Next;`
Same comment about linked lists.

`cmsUInt16Number **White, vs. const cmsUInt16Number **White,`

This is exported by the API, I cannot change it, 

So, I opted to get only some of your changes, which are nice finding, so thank you again for the review and the PR!

Best regards
Marti

PR #509, opened 2025-08-22
Fix adaptation issues when mixing V2 and V4 profiles in the pipeline

dimula73:
LCMS handles white point adaptation differently, depending on whether the profile's white point is defined via cmsSigChromaticAdaptationTag or cmsSigMediaWhitePointTag. If the white point is set via cmsSigMediaWhitePointTag, LCMS uses "XYZ Scaling" method to adapt the colors. If the white point is defined via cmsSigChromaticAdaptationTag, then LCMS uses the embedded matrix (which is most likely "Bradford").

It causes some weird issues when combining profiles with white point defined differently. For example, when color proofing a D65 device on a D65 screen with absolute colorimetric intent, the D65->D50->D65 conversion may cause severe color degradations (instead of being identity).

The actual testcase looks like that:

* Proof profile: ICCv2, RGB "output" with D65 white point set via cmsSigMediaWhitePointTag

* Display profile: ICCv2, sRGB "display" with D65 white point set via cmsSigChromaticAdaptationTag (e.g. Elle's sRGB-v2 profile)

When combining these profiles in the pipeline, the conversions look like that:

1) AToB1 is applied for the proof profile (D65 adaptation is
   converted into D50)
2) ComputeAbsoluteIntent adapts back D50 to D65 using "XYZ Scaling"
   method (because the output profile has only cmsSigMediaWhitePointTag)
3) ComputeAbsoluteIntent also applies also inverse adaptation of the
   display profile, which is "Bradfort", since the profile has
   cmsSigChromaticAdaptationTag.
4) BToA1 is applied for the display profile

In general, we convert a D65 space into D65 space with absolute intent, so, theoretically, we expect steps 2 and 3 result in a noop. But, given that the two profiles used a different adaptation method, the result is not a noop and skews the colors into a bluish tint.

It looks like someone has already seen this issue with display class profiles (more than 15 years ago) and added a special case for them, which forces Bradfort matrix for cmsSigMediaWhitePointTag-only profiles.

This patch just extends this hack to the output-class profiles. Their cmsSigMediaWhitePointTag is effectively converted into cmsSigChromaticAdaptationTag with a Bradfort matrix.

PS:
This patch might a bigger discussion on how cmsSigChromaticAdaptationTag should be handled. What happens if the user has two profiles with two cmsSigChromaticAdaptationTag tags that use different adaptation methods? Should CMS just discard the matrix and force the homogeneous method for all the profiles?

PPS:
The patch also adds a unittest to verify the issue. Though I'm not fully sure if 'output-v2-d65-no-chad.icc' is correct from ICC specification point of view. I just used a modified version of cmsCreateRGBProfile() to generate it. I wanted all the LUTs of the profiles to be exactly the same to make the test possible.


mm2:
Thanks for the PR.

I think, however, you are mixing media white point with illuminant. White point is the chromaticity of unprinted paper for print profiles; illuminant is the light source used to see and measure the printed colors.

ICC clearly states that display profiles must have D50 as media white (see https://color.org/whitepapers/ICC_White_Paper_6_v2_and_v4_display_profile_differences-updated.pdf) So, **the profile "display-v2-d65-no-chad.icc" is wrong**. Little CMS has code to specifically fix those wrong profiles by forcing the media white to D50, this is mostly because the old sRGB profile was affected by this bug. There was a discussion in the ICC committee, where I was a company representative years ago. The amount of those "hot fixes" in Little CMS code is minimum, but the sRGB bug was affecting such amount of workflows that we decide to fix it on the CMM.

The CHAD tag is used to recover absolute colorimetry, in the CIE-absolute sense, when observer is unadapted. That means, taking the light source into account. The media white tag is used to recover unprinted paper white when INTENT_ABSOLUTE_COLORIMETRIC intent is used.  See cmsSetAdaptationState() for turning on this feature.

**The modification you propose would break all absolute colorimetric intents on all V2 printers**.  Media white tag holds exactly that, the unprinted media color. 

Regarding adaptation methods, ICC uses same chromatic adaptation as CIE Lab to handle paper discounting. The main difference is doing the Von-Kries scaling in XYZ or in the LMS cone space. In real world the so called "Wrong-von-kries" works well in most cases.

About the other profiles you use:

Proof profile: ICCv2, RGB "output" with D65 white point   set via cmsSigMediaWhitePointTag --> that is correct, this is a RGB printer with a paper that has same chromaticity as a D65 light, so is bluish if seen under D50.

Display profile: ICCv2, sRGB "display" with D65 white point set via  cmsSigChromaticAdaptationTag (e.g. Elle's sRGB-v2 profile) --> That is almost right, white point and CHAD are ok, but it misses perceptual black scaling. lcms handle that, though.

CheckProofingXFORM16_v2_output_no_chad: You are simulating how a patch printer on a bluish paper (chromaticity of d65) behaves. The light you use is D65, but you are fully adapted to this light. 

I have reviewed the tests and **without your changes** you got 1 level of difference out of 63535. You almost compensate chromaticities but not entirely because you are comparing different things. That is a quite reasonable result.  For second test, you are using a wrong profile, so results may be any. But even with the wrong profile, the test passes ok. 

**TL;DR:** you cannot mix the light using to illuminate the paper with the color of the paper. Display profiles use a virtual paper that is always D50. You are using a wrong profile for testing.

mm2:
> Sorry, but I cannot agree with this sentence.

That's the very first paragraph in the linked document. ICC is normative on that. v2 was not well defined and with v4 they tried to clarify.

There was a huge discussion on the ICC about all this, 15 or 16 years ago. It is pointless to discuss it again, and I have no way to fix   what ICC have decided. 

> As a proofing space select any RGB printer ICCv2 output profile that has cmsSigMediaWhitePointTag set to D65 and doesn't have any cmsSigChromaticAdaptationTag tag
> Select "Absolute Colorimetric" as the display intent (not the proofing intent)
> You will see that the color of the image are skewed into bluish side, even though the white point of the source and destination color spaces is the same, i.e. no adjustment should happen.

That is correct, if printer has media white point D65, the paper is blue. Monitors do nothing on ICC absolute intent. Monitors have D50 as media white. You can select absolute or relative for display and result is same. You are assumed to be adapted to the light.  This is how ICC works. You can like it or not. I don't like either, so I added a function to set the adaptation state of the observer. But again, what you describe is the expected behavior following ICC rules. If you want the light of the monitor to be taken into account, use cmsSetAdaptionState(0)

dimula73:
> There was a huge discussion on the ICC about all this, 15 or 16 years ago. It is pointless to discuss it again, and I have no way to fix what ICC have decided.

Ok, well, that particular profile in question was to unittest the previous hack. There is not much use in discussing whether it is correct or not. It just exists in the wild :)

> That is correct, if printer has media white point D65, the paper is blue

Do I understand it right that `cmsSigMediaWhitePointTag` is the color of the paper and `cmsSigChromaticAdaptationTag` is the result of the viewing conditions the eye is adapted to?

Which means that the profile with D65 set as `cmsSigMediaWhitePointTag` and no `cmsSigChromaticAdaptationTag` means that that is a blueish paper viewed under D50 environment? 

If so, then I return to my original "bug" that I got while implementing Wayland color management support in Krita. When the display profile is set to a D50 profile (as reported by Wayland), the "adaptation" slider does nothing. Do I understand it right that the adaptation slider works only for adapting the "viewing conditions" and not the paper white? I personally expected that it would adapt to the paper white as well...

If the last statement is correct, is it possible to add proper documentation for that AdaptationState somewhere? Is there any kind of repo for the docs? I had really hard time understanding what it was supposed to do.

mm2:
> Do I understand it right that cmsSigMediaWhitePointTag is the color of the paper and cmsSigChromaticAdaptationTag is the result of the viewing conditions the eye is adapted to?

Yes, more precisely cmsSigChromaticAdaptationTag is the chromatic adaptation of your eye to the light.

> Which means that the profile with D65 set as cmsSigMediaWhitePointTag and no cmsSigChromaticAdaptationTag means that that is a blueish paper viewed under D50 environment?

Right.

> When the display profile is set to a D50 profile (as reported by Wayland), the "adaptation" slider (the one driving cmsSetAdaptation()) does nothing. 

For this to work you need to use INTENT_ABSOLUTE_COLORIMETRIC and use a profile with cmsSigChromaticAdaptationTag.
You can try transicc:

```
marti@I7:~$ transicc -d1 -t3 -i*sRGB -o*Lab
R? 180
G? 180
B? 180
L*=73.3120 a*=-0.0000 b*=0.0000

marti@I7:~$ transicc -d0 -t3 -i*sRGB -o*Lab
R? 180
G? 180
B? 180
L*=73.3120 a*=-1.8379 b*=-14.9405

marti@I7:~$ transicc -d0.5 -t3 -i*sRGB -o*Lab
R? 180
G? 180
B? 180
L*=73.2729 a*=-0.8220 b*=-6.3438
```
>If the last statement is correct, is it possible to add proper documentation for that AdaptationState somewhere? 
I put it high in the TODO list. Little CMS handles incomplete adaptation states as a hidden gem, I know of several softproof programs using this feature.

dimula73:
> For this to work you need to use INTENT_ABSOLUTE_COLORIMETRIC and use a profile with cmsSigChromaticAdaptationTag.

Yes, in my case the target profile was generated with lcmsCreateRGBProfile and had an adaptation tag, but it was almost identity, because it was "almost D50" (due to rounding errors)

> I put it high in the TODO list. Little CMS handles incomplete adaptation states as a hidden gem, I know of several softproof programs using this feature.

Well, Krita uses this slider in its Color Proofing feature, that is how I found the issue. Can we write the draft text and send somewhere?

dimula73:
Hi, @mm2!

Just a random question (related to Wayland protocol specification): can "ColorSpace"-class profiles have `cmsSigMediaWhitePointTag` and should they be adapted when passing through absolute-colorimetric transformation?

mm2:
> Hi, @mm2!
> 
> Just a random question (related to Wayland protocol specification): can "ColorSpace"-class profiles have `cmsSigMediaWhitePointTag` and should they be adapted when passing through absolute-colorimetric transformation?

Since this is "media white" it should be handled by using the "wrong-von-kries" method that ICC and Lab use. Thats's media color, as name suggest. Most of those spaces have D50, though.

dimula73:
> Since this is "media white" it should be handled by using the "wrong-von-kries" method that ICC and Lab use. Thats's media color, as name suggest. Most of those spaces have D50, though.

But is it considered "sane" for a "color space" profile to have "media white" at all?

PR #528, opened 2025-12-30
Discount Relative+BPC ink limiting for CMYK and cmsSigNcolorData color spaces

ryanobeirne:
Hi! I am working in the digital textile printing space where `CMYK+N` printer profiles are common (`CMYK+Orange+Violet`, `CMYK+Red+Green+Blue`, etc.). I've run into an issue where lcms fails to detect the black point for the Relative Colorimetric intent on `CMYK+N` ICC profiles.

For example, converting the colors `RGB8(0,0,0)` (black) and `RGB8(36,36,36)` (dark gray) from `sRGB` to a `CMYK+3` profile using `Relative+BPC` should produce 2 distinct grays, but lcms outputs the same color for both samples. This is in contrast to the Adobe BPC implementation in Photoshop that does the same transform but produces 2 distinct grays.

I believe I have traced the issue to the `cmsDetectBlackPoint` function in `cmssamp.c`: it appears as though lcms currently uses the Perceptual black point for `Relative+BPC` transforms _only on CMYK profiles_. I believe the Perceptual black point should also be used for `Relative+BPC` transforms on `CMYK+N` profiles as well.

In this patch, I have extended the `cmsDetectBlackPoint` function to use the Perceptual black point for `cmsSigCmykData` color spaces _AND_ all the `cmsSigNcolorData` color spaces. However, I'm not sure if _all_ of these color spaces should be included or if any additional color spaces should be included as well.

I have tested this patch on `cmsSig6-8colorData` output profiles (`CMYK+2`..`CMYK+4`) and found the results to be congruent with Adobe Photoshop. I am happy to provide `CMYK+N` profiles for testing if it is helpful.

### DIFF
--- a/include/lcms2.h
+++ b/include/lcms2.h
@@ -1570,6 +1570,9 @@ CMSAPI cmsUInt32Number   CMSEXPORT cmsChannelsOf(cmsColorSpaceSignature ColorSpa
 // Get number of channels of color space or -1 if color space is not listed/supported
 CMSAPI cmsInt32Number CMSEXPORT cmsChannelsOfColorSpace(cmsColorSpaceSignature ColorSpace);
 
+// Determine if Relative renderings with Black Point Compensation should discount ink limiting for the color space
+CMSAPI cmsBool CMSEXPORT cmsDiscountRelativeInkLimiting(cmsColorSpaceSignature ColorSpace);
+
 // Build a suitable formatter for the colorspace of this profile. nBytes=1 means 8 bits, nBytes=2 means 16 bits. 
 CMSAPI cmsUInt32Number   CMSEXPORT cmsFormatterForColorspaceOfProfile(cmsHPROFILE hProfile, cmsUInt32Number nBytes, cmsBool lIsFloat);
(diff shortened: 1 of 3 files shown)


ryanobeirne:
I assume you refer to this line in `adobebpc.pdf`:

> Although ICC profiles can use generalized n-Color spaces, Adobe Systems currently does not apply its BPC
algorithm to such color spaces

It would appear this is no longer the case (nor should it ever have been, in my opinion). Adobe is definitely using BPC when converting to `Multichannel` profiles when the BPC option is selected in Photoshop.

multichannel-bpc

Additionally, I know of at least one RIP using LCMS as the color engine that similarly patches LCMS to do the same.

I still question my decision here to use `cmsSigCmykData` and all `cmsSigNcolorData` color spaces as the criteria for selecting the perceptual black point. Should the criteria be that the profile contains `CMYK` with or without additional channels? Does the distinction matter?

mm2:
I have just released 2.18 and now I am ready to check it. I really appreciate your aid with this. In fact, BPC is just a plain scaling in XYZ and the difficult part is detecting black. Since there are many blacks and profiles behave differently I have to check what multi channel profiles do in such cases. I am keeping this PR open for the investigation.

PR #531, opened 2026-01-08
CMake build system for lcms2

ssh4net:
CMake-based build system for lcms2, 
including options for building shared/static libraries, tools, tests, and optional plugins. 
The new CMake scripts handle feature detection, packaging, installation, and integration with external dependencies such as JPEG, TIFF, and ZLIB.

Tested on MacOS and Windows (static only).
For a static build of libtiff, there should be Find***.cmake for all 3rd party dependencies; otherwise, CMake can't have a clue what dependencies tiff was built with. Native libtiff does not have such features. This is an issue from libtiff side and should be fixed there.


mm2:
I have **fully tested** this PR on about 10 assorted computers and OS and works great. Excellent job, thank you so much. I am going to add you as contributor. Please let me know if that is ok to you and if you want to be listed as Vlad Erium.
Many thanks again, see the merge in a short!

PR #555, opened 2026-04-19
Fix PostScript injection via ICC ncl2 named colors (#554)

### LINKED ISSUE #554: [Bug] High: PostScript injection via ICC ncl2 named colors
Priority: High

Affected component:
- PostScript CSA / CRD generation in `cmsGetPostScriptCSA()` and `cmsGetPostScriptCRD()`
- Reachable from the shipped `psicc` CLI and from API consumers that emit PostScript color resources

Verified on:
- `fb48851f48f505fc938312dbc2dbe1cab56da7ec` (`master`)

Root cause:
`Type_NamedColor_Read()` accepts `ncl2` named-color entries from the ICC profile and stores the 32-byte root name. `WriteNamedColorCSA()` and `WriteNamedColorCRD()` later emit that name directly into PostScript string literals with `_cmsIOPrintf(m, "  (%s) ...", ColorName, ...)` without escaping `(`, `)`, or `\\`. A crafted name such as `foo) /exec (bar` terminates the intended string and injects raw PostScript tokens into the generated output.

Exact reproduction steps:
1. Run the direct API-backed verifier:
   ```bash
   ASAN_OPTIONS=detect_leaks=0 LD_LIBRARY_PATH=$PWD/build-asan \
     build-asan/verify/t2_namedcolor_ps_repro \
     tests/regression/t2-namedcolor/t2-ncl2-ps-injection.icc
   ```
2. Generate PostScript through the shipped CLI:
   ```bash
   ASAN_OPTIONS=detect_leaks=0 LD_LIBRARY_PATH=$PWD/build-asan \
     build-asan/psicc -itests/regression/t2-namedcolor/t2-ncl2-ps-injection.icc > /tmp/t2-csa.ps

   ASAN_OPTIONS=detect_leaks=0 LD_LIBRARY_PATH=$PWD/build-asan \
     build-asan/psicc -otests/regression/t2-namedcolor/t2-ncl2-ps-injection.icc > /tmp/t2-crd.ps
   ```
3. Confirm the raw injected token appears in both outputs:
   ```bash
   grep -n 'foo) /exec (bar' /tmp/t2-csa.ps /tmp/t2-crd.ps
   ```
4. Observed verification output includes:
   - `CSA raw named-color token: foo) /exec (bar)`
   - `CRD raw named-color token: foo) /exec (bar)`
   - The generated PostScript contains lines such as `(foo) /exec (bar) [ ... ]`

Verification evidence:
- The verification stage added a dedicated named-color profile generator and API-backed oracle.
- The bug reproduced through both the direct verifier and the shipped `psicc` CLI on CSA and CRD output paths.
- The focused semantic fuzzer also replayed the malicious seed and emitted an identical crash artifact.

Relevant file paths:
- `src/cmstypes.c:3405-3411`
- `src/cmsps2.c:1018-1022`
- `src/cmsps2.c:1452-1457`
- `utils/psicc/psicc.c:157-166`
- `utils/psicc/psicc.c:200-208`
- `tests/regression/t2-namedcolor/t2_namedcolor_ps_repro.c`
- `tests/regression/t2-namedcolor/t2-ncl2-ps-injection.icc`

Suggested fix direction:
- Add a dedicated PostScript string-literal escaping helper that emits `\\`, `\(`, and `\)` for special characters and octal escapes for non-printable bytes.
- Apply that helper to named-color emission in both CSA and CRD paths.
- Audit other `_cmsIOPrintf("(%s)", ...)` sites in `src/cmsps2.c` so all PostScript literal emitters consistently escape attacker-controlled text.

parasol-aser:
## Summary

Fixes #554 — High-severity PostScript injection in `cmsGetPostScriptCSA()` / `cmsGetPostScriptCRD()`.

`Type_NamedColor_Read()` accepts the 32-byte `Root` color name from an ICC profile's `ncl2` tag verbatim. `WriteNamedColorCSA()` (`src/cmsps2.c:1022`) and `WriteNamedColorCRD()` (`src/cmsps2.c:1457`) later emit that name directly into a PostScript string literal via `_cmsIOPrintf(m, "  (%s) ...", ColorName, ...)` with **no escaping** of `(`, `)`, or `\`. A crafted name such as `foo) /exec (bar` closes the intended literal and injects raw PostScript tokens into the generated output consumed by downstream RIPs (Ghostscript, printer RIPs), yielding code execution or data exfiltration on the consumer.

### What changed

- **`src/cmsps2.c`**: new static helper `EmitPSEscaped()` near `RemoveCR()` writes a PS string-literal body with PLRM §3.3.4.1-compatible escaping:
  - `\`, `(`, `)` → `\\`, `\(`, `\)`
  - bytes `< 0x20` or `>= 0x7F` → three-digit `\ddd` octal
  - other printable ASCII passes through unchanged
- `WriteNamedColorCSA()` and `WriteNamedColorCRD()` now route `ColorName` through `EmitPSEscaped()`. The surrounding indentation (`  (`) and trailing ` [ ... ]\n` are preserved verbatim so legitimate profile output differs only in the escaped bytes themselves.
- **`testbed/testcms2.c`**: adds `CheckNamedColorPSEscaping()`, registered alongside `CheckPostScript`, exercising the issue's reproducer across 15 scenarios.

### Scope rationale

The fix is **output-encoding at the sink**, not input validation. ICC permits arbitrary 7-bit-ASCII bytes in `Root`, so rejecting them at parse time would break legitimate profiles and would not fix API consumers that build named-color lists in memory via `cmsAppendNamedColor()`.

Audit of other `_cmsIOPrintf("(%s)", ...)` sites in `cmsps2.c` (per the reporter's suggestion): only the two named-color sinks above carry attacker-controlled text. Comment emitters (`%% %s\n`) use `RemoveCR()`, `EmitIntent()` uses a hardcoded `switch`, and the `Prefix`/`Suffix` arrays are hardcoded literals today — no additional sinks to escape.

### Reproducer (from the issue)

Before fix:

```bash
ASAN_OPTIONS=detect_leaks=0 LD_LIBRARY_PATH=$PWD/build-asan \
  build-asan/psicc -i tests/regression/t2-namedcolor/t2-ncl2-ps-injection.icc > /tmp/t2-csa.ps
grep 'foo) /exec (bar' /tmp/t2-csa.ps   # matches — raw injection present
```

After fix: the grep finds only the escaped form `foo\) /exec \(bar` inside the `(...)` literal.

## Test plan

- [x] `CheckNamedColorPSEscaping` added to `testbed/testcms2.c`, covering:
  - Primary `foo) /exec (bar\baz\x01\xff` injection payload — raw absent, fully escaped form present in both CSA and CRD
  - Benign `Hazelnut 14-1315` and `PANTONE Orange 021 C` — pass through verbatim, no over-eager escaping
  - Empty name → `  ()`
  - Standalone backslash → `\\`
  - All-open-paren and all-close-paren names (2× expansion)
  - Boundary bytes around the printable range: `0x1F`, `0x20`, `0x7E`, `0x7F`
  - Raw high-bit bytes (`0xFF`) → `\377` (4× expansion)
  - `LF`/`CR`/`TAB` → `\012`/`\015`/`\011`
  - Mixed printable + metacharacter interleave
  - Stress: 31 `(` bytes (ncl2 Root upper bound) → 62 escape bytes
  - Trailing backslash does not escape the closing `)`
  - Single-`)` minimal repro
- [x] `testcms` test suite runs clean via CMake build (`ctest -R testcms`) — all existing tests still pass, including the prior `CheckPostScript` suite that exercises non-named-color profiles.
- [x] Manual verification with the issue's reproducer: `psicc -i` / `psicc -o` on `t2-ncl2-ps-injection.icc` no longer emits `foo) /exec (bar`; the escaped form `foo\) /exec \(bar` appears instead.
- [ ] CI on this PR (CMake and Meson matrices) passes.

### DIFF
--- a/src/cmsps2.c
+++ b/src/cmsps2.c
@@ -329,6 +329,30 @@ char* RemoveCR(const char* txt)
 
 }
 
+// Writes the body of a PostScript string literal, escaping the metacharacters
+// '\\', '(' and ')' and emitting non-printable / high-bit bytes as octal
+// triples per PLRM 3.3.4.1. The caller is responsible for the surrounding
+// '(' and ')' delimiters.
+static
+void EmitPSEscaped(cmsIOHANDLER* m, const char* txt)
+{
+    const unsigned char* p;
+
+    if (txt == NULL) return;
+
+    for (p = (const unsigned char*) txt; *p != 0; p++) {
+        unsigned char c = *p;
+
+        if (c == '\\' || c == '(' || c == ')') {
+            _cmsIOPrintf(m, "\\%c", c);
+        } else if (c < 0x20 || c >= 0x7F) {
+            _cmsIOPrintf(m, "\\%03o", c);
+        } else {
+            _cmsIOPrintf(m, "%c", c);
+        }
+    }
+}
+
 static
 void EmitHeader(cmsIOHANDLER* m, const char* Title, cmsHPROFILE hProfile)
 {
@@ -1019,7 +1043,9 @@ int WriteNamedColorCSA(cmsIOHANDLER* m, cmsHPROFILE hNamedColor, cmsUInt32Number
                 continue;
 
         cmsDoTransform(xform, In, &Lab, 1);
-        _cmsIOPrintf(m, "  (%s) [ %.3f %.3f %.3f ]\n", ColorName, Lab.L, Lab.a, Lab.b);
+        _cmsIOPrintf(m, "  (");
+        EmitPSEscaped(m, ColorName);
+        _cmsIOPrintf(m, ") [ %.3f %.3f %.3f ]\n", Lab.L, Lab.a, Lab.b);
     }
 
     _cmsIOPrintf(m, ">>\n");
@@ -1454,7 +1480,9 @@ int WriteNamedColorCRD(cmsIOHANDLER* m, cmsHPROFILE hNamedColor, cmsUInt32Number
 
         cmsDoTransform(xform, In, Out, 1);
         BuildColorantList(Colorant, nColorant, Out);
-        _cmsIOPrintf(m, "  (%s) [ %s ]\n", ColorName, Colorant);
+        _cmsIOPrintf(m, "  (");
+        EmitPSEscaped(m, ColorName);
+        _cmsIOPrintf(m, ") [ %s ]\n", Colorant);
     }
 
     _cmsIOPrintf(m, "   >>");
--- a/testbed/testcms2.c
+++ b/testbed/testcms2.c
@@ -7199,6 +7199,452 @@ cmsInt32Number CheckPostScript(void)
     return 1;
 }
 
+// ---------------------------------------------------------------------------
+// Named-color PostScript-literal escaping (issue #554 / T-2).
+//
+// The plan: name bytes passed to cmsGetPostScriptCSA/CRD for a named-color
+// profile must be emitted inside a PS literal string with ( ) \ escaped and
+// bytes <0x20 or >=0x7F emitted as three-digit \ddd octal.  Raw injection
+// payloads like "foo) /exec (bar" must never appear unescaped.
+// ---------------------------------------------------------------------------
+
+static
+cmsHPROFILE BuildNamedColorProfile(const char* name)
+{
+    cmsHPROFILE placeholder = NULL;
+    cmsHPROFILE loaded = NULL;
+    cmsNAMEDCOLORLIST* colors = NULL;
+    cmsMLU* description = NULL;
+    cmsMLU* copyright = NULL;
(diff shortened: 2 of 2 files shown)


mm2:
Add the fix in c22dc8623fd10a0571362a1dc91a3fe8e29496e7

mm2:
@parasol-aser  thanks for the fix. Since this is on an undocumented function -and I want to keep it as such-, I would prefer to not include it on the testbed.

parasol-aser:
> @parasol-aser thanks for the fix. Since this is on an undocumented function -and I want to keep it as such-, I would prefer to not include it on the testbed.

@mm2 that looks fine to me. thank you!

PR #567, opened 2026-05-06
recover the undocumented feature needed by colord (again)

xry111:
Use the fix suggested by the reporter of #542 instead of erroring out to bring back the undocumented feature required by colord.


mm2:
Thanks, I will merge it as soon as the checks complete

PR #575, opened 2026-05-30
Fix compilation on old MSVC

jankratochvil:
cmscgats.c(3234) : error C2275: 'cmsFloat32Number' : illegal use of this type as an expression


mm2:
Thanks that was from one of those “security” fixes, I should  do more strict code reviews.

PR #587, opened 2026-07-18
Fix write side of wchar_t <-> UTF-16 conversion for supplementary-plane code points

### LINKED ISSUE #180: Misuse of wchar_t for UTF-16/32
Even though `wchar_t` does not have a well-defined size in C or C++, LittleCMS uses it to store UTF-16 strings.

The problem is that on platforms with a 32-bit `wchar_t` (e.g. Linux, OS X), each UTF-16 code unit (e.g. from a file) is simply cast to `wchar_t` by `_cmsReadWCharArray` which doesn't produce a valid UTF-32 sequence. To produce a valid UTF-32 sequence one would have to decode each Unicode code point from one or more UTF-16 code units.

To make things worse `_cmsWriteWCharArray` just down-casts `wchar_t` to `cmsUInt16Number` discarding the upper 16 bits of the UTF-32 code unit. This of course does not convert UTF-32 to UTF-16, as the correct conversion would be to decompose each Unicode code point into one or more UTF-16 code units.

Apart from these problems there is also a minor matter of wasted memory - each 32-bit `wchar_t` has 16 bits unused.

At the very minimum, I would propose documenting the fact that strings created by LittleCMS functions returning `wchar_t` arrays cannot be used as valid UTF-32 strings on platforms with a 32-bit `wchar_t`. Instead the user must do their own conversion, using the lower 16 bits of each value to form a valid UTF-16 sequence. And functions that take `wchar_t` strings should also be documented to take UTF-16 sequences where each code unit is stored in a `wchar_t`.

Of course, this behaviour is not intuitive or expected by the user and makes handling textual data with LittleCMS more complex (not to mention resource intensive, as each string conversion may require allocating extra memory). For instance, this confusion can cause issues such as #163.

That is why I also propose deprecating the functions that use `wchar_t` in favour of new functions that operate on some new `cmsU16Character` type (always 16-bit, name definitely up for discussion), and explicitly state that strings are encoded as UTF-16 in the native byte order. (I would say, just use `char16_t` but that would require C11/C++11, so a custom type is probably a better choice for compatibility). The new functions could use a `UTF16` postfix.
With these new functions the internals would need to be reworked. Instead of storing Unicode text as `wchar_t` arrays, it should be stored as `cmsU16Character` arrays. As an added bonus, this would allow for more efficient I/O functions that could process multiple characters at once instead of a character at a time (with the byte-swapping `if`-statement needing to be called only once for character block).
The deprecated functions would just do the up/down-casting themselves instead of deferring that to the I/O layer.

Another issue that I see is with the `ASCII` functions as the conversion from UTF-16 to ASCII (`cmsMLUgetASCII`) is only valid if the original string has no non-ASCII code points in it. A better way to handle this may be to provide the ability to replace non-ASCII code points with a character of the user's choosing (e.g. `?`). Of course this would require handling strings as actual sequences of code points instead of opaque code units. (I'm not 100% sure, but this probably doesn't even require fully decoding the UTF-32 code point from UTF-16, just skipping over some of the UTF-16 code units.)

P.S. I would be glad to create a pull-request with some of the proposed changes if a decision is made to proceed with them.

hmms:
Fixes #180.

## The problem

Little-CMS stores ICC profile description/copyright strings (`desc` and
`mluc` tags) internally as `wchar_t` arrays, one Unicode code point per
element. On platforms where `wchar_t` is 32 bits (Linux/macOS), any code
point fits in a single `wchar_t`, including supplementary-plane characters
above U+FFFF.

The on-disk ICC format for these tags is always UTF-16, where a code point
above U+FFFF has to be split into a two-unit surrogate pair. The read side
already handles this correctly (2.13 added `convert_utf16_to_utf32`), but as
@mm2 noted in the issue, the write side never did — `_cmsWriteWCharArray`
just truncated every value to 16 bits:

```c
// before
for (i=0; i < n; i++) {
    if (!_cmsWriteUInt16Number(io, (cmsUInt16Number) Array[i])) return FALSE;
}
```

Concretely, U+1F600 (😀) should serialize as the two-unit surrogate pair
`D8 3D DE 00`; before this fix it serialized as the single (wrong, and
mangled by the truncation) unit `E6 00`.

## Why this touches more than `_cmsWriteWCharArray`

Fixing just that loop isn't safe by itself: encoding a surrogate pair means
writing *more* on-disk bytes than the `wchar_t` count would predict, and two
tag writers precompute exact sizes/offsets from that count before writing
any data:

- `Type_Text_Description_Write` (legacy `desc` tag) declared the on-disk
  Unicode length field as exactly the ASCII length, and sized the whole tag
  from that same value.
- `Type_MLU_Write` (modern `mluc` tag) computed each entry's on-disk
  `Len`/`Offset` position-table fields via a fixed 1:1 scale between
  `wchar_t` count and on-disk units.

Both needed to account for the possible extra unit per surrogate pair, or a
surrogate pair anywhere in the string would desync the tag's declared size
or a later entry's offset from what's actually written, corrupting the tag.

While fixing `Type_MLU_Write`'s position table, I found an independent,
pre-existing bug on the *read* side (`Type_MLU_Read`): it converts on-disk
byte offsets into decoded-buffer offsets via the same kind of fixed 1:1
scale, computed *before* decoding — wrong whenever an earlier entry (in
on-disk order) contains a surrogate pair, since two on-disk units collapse
into one decoded `wchar_t`. A correct write side would be useless if nothing
it wrote could be read back correctly, so this needed fixing too.

## Summary of changes (`src/cmstypes.c`)

- `convert_utf32_to_utf16` (new): write-side mirror of the existing
  `convert_utf16_to_utf32`, encoding code points above U+FFFF as UTF-16
  surrogate pairs. Wired into `_cmsWriteWCharArray`, gated on
  `sizeof(wchar_t) > sizeof(cmsUInt16Number)` so 16-bit-`wchar_t` platforms
  (Windows) are unaffected.
- `_cmsUtf16UnitsForScalar` / `_cmsCountUtf16Units`: shared classification of
  a code point into "1 unit," "2 units," or "not a valid Unicode scalar"
  (values above U+10FFFF, or a lone surrogate code point U+D800–U+DFFF used
  directly as a scalar). Used by every site that needs to know a value's
  on-disk size, so counting and encoding can't disagree, and an unencodable
  value now makes the write fail cleanly instead of emitting structurally
  invalid UTF-16.
- `Type_Text_Description_Write`: computes the real on-disk unit count up
  front and uses it for both the declared length field and the tag-size
  calculation.
- `Type_MLU_Write`: builds a whole-pool position map (decoded `wchar_t`
  index → on-disk unit offset) once, then looks up each entry's
  `Offset`/`Len` through it. This also correctly preserves entries that
  legitimately share one on-disk string (some encoders dedup this way) — a
  naive running-offset accumulator (an earlier draft of this fix) breaks
  that case, and also validates each entry's range against the pool before
  indexing the map.
- `Type_MLU_Read`: added `convert_utf16_to_utf32_indexed`, which decodes the
  pool once while building a map from on-disk unit index to decoded
  `wchar_t` index, so each entry's `StrW`/`Len` can be resolved correctly
  after decoding instead of via the old pre-decode fixed scale. The map is
  sentinel-filled so a position landing in the middle of a surrogate pair
  (a valid byte offset, but not a valid character boundary) is explicitly
  rejected rather than silently treated as index 0, which could otherwise
  underflow the computed length.

## Testing

Added 4 regression tests to `testbed/testcms2.c` (`CheckMLU_SupplementaryPlane`,
`CheckMLU_MalformedSurrogateOffset`, `CheckMLU_SharedEntryRoundtrip`,
`CheckMLU_ScalarValidation`), covering: round-tripping a supplementary-plane
character through both `desc` and `mluc` (with the character placed before
another entry, to exercise the position-table fix); a hand-built malformed
`mluc` tag with an entry offset landing mid-surrogate-pair, confirming clean
rejection; two entries sharing one on-disk string, confirming the sharing
survives a read → write → read cycle; and boundary/invalid scalar values
(U+FFFF, U+10000, U+10FFFF valid; a lone surrogate and a value above
U+10FFFF invalid).

Full existing test suite passes with no regressions. Re-ran everything under
AddressSanitizer and macOS's `leaks` tool — both clean (0 errors, 0 leaks).

Given this touches ICC tag parsing/serialization, I tried to be conservative
and verify each change concretely rather than just by inspection — several
of the fixes above (the `Type_MLU_Read` bug, a memory leak on one of the new
error paths, and the shared-entry regression in an earlier draft of the
`Type_MLU_Write` fix) were caught this way, each confirmed by constructing a
concrete failing case and reproducing it before fixing.

No `wchar_t` used as a genuine `char16_t` on Windows should see any
behavior change — the new code paths are all gated on
`sizeof(wchar_t) > sizeof(cmsUInt16Number)`.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 -----------------------
 2.20 Ongoing
 -----------------------
+Fix write side of wchar_t <-> UTF-16 conversion on platforms with 32-bit wchar_t, issue #180
 
 -----------------------
 2.19.1 hot fix
--- a/src/cmstypes.c
+++ b/src/cmstypes.c
@@ -1784,7 +1784,7 @@ void *Type_MLU_Read(struct _cms_typehandler_struct* self, cmsIOHANDLER* io, cmsU
     cmsUInt32Number SizeOfHeader;
     cmsUInt32Number  Len, Offset;
     cmsUInt32Number  i;
-    wchar_t*         Block;
+    wchar_t*         Block = NULL;
     cmsUInt32Number  BeginOfThisString, EndOfThisString, LargestPosition;
     cmsUInt32Number* RawBegin = NULL;   // Per-entry on-disk begin, in UTF-16 code units
     cmsUInt32Number* RawLen = NULL;     // Per-entry on-disk length, in UTF-16 code units
@@ -1872,16 +1872,11 @@ void *Type_MLU_Read(struct _cms_typehandler_struct* self, cmsIOHANDLER* io, cmsU
         // with an invalid marker up front (instead of calloc's 0) makes that gap
         // distinguishable from a real, valid index 0 below.
         UnitToWChar = (cmsUInt32Number*) _cmsMalloc(self ->ContextID, (NumOfWchar + 1) * sizeof(cmsUInt32Number));
-        if (UnitToWChar == NULL) {
-            _cmsFree(self->ContextID, Block);
-            goto Error;
-        }
+        if (UnitToWChar == NULL) goto Error;
+
         memset(UnitToWChar, 0xff, (NumOfWchar + 1) * sizeof(cmsUInt32Number));
 
-        if (!convert_utf16_to_utf32_indexed(io, NumOfWchar, Block, UnitToWChar)) {
-            _cmsFree(self->ContextID, Block);
-            goto Error;
-        }
+        if (!convert_utf16_to_utf32_indexed(io, NumOfWchar, Block, UnitToWChar)) goto Error;
     }
 
     // Translate each entry's on-disk position into indices into the decoded Block,
@@ -1928,6 +1923,7 @@ void *Type_MLU_Read(struct _cms_typehandler_struct* self, cmsIOHANDLER* io, cmsU
     return (void*) mlu;
 
 Error:
+    if (Block) _cmsFree(self ->ContextID, Block);
     if (RawBegin) _cmsFree(self ->ContextID, RawBegin);
     if (RawLen) _cmsFree(self ->ContextID, RawLen);
     if (UnitToWChar) _cmsFree(self ->ContextID, UnitToWChar);
--- a/src/cmstypes.c
+++ b/src/cmstypes.c
@@ -128,10 +128,31 @@ cmsINLINE cmsUInt32Number surrogate_to_utf32(cmsUInt32Number high, cmsUInt32Numb
     return (high << 10) + low - 0x35fdc00;
 }
 
-// Number of on-disk UTF-16 code units required to encode n wchar_t elements.
-// On platforms where wchar_t is already 16 bits this is always n. On 32-bit
-// wchar_t platforms, any code point above 0xffff needs a UTF-16 surrogate
-// pair (two 16-bit units) to be represented on disk.
+// Sentinel returned by _cmsUtf16UnitsForScalar for a value that cannot be encoded as
+// a single valid UTF-16 scalar value on disk.
+#define _cmsUTF16_INVALID_SCALAR ((cmsUInt32Number) 0xffffffffu)
+
+// Classify a single code point for on-disk UTF-16 encoding: returns the number of
+// UTF-16 code units it needs (1 for the BMP, 2 for a surrogate pair), or the sentinel
+// above if the value is not representable at all -- either above the maximum valid
+// Unicode scalar value (U+10FFFF), or itself a surrogate code point (U+D800-U+DFFF),
+// which can only appear as one half of an on-disk pair, never as a scalar value in its
+// own right. Both counting and encoding go through this single classification so they
+// can never disagree with each other about how many units a given value takes.
+cmsINLINE cmsUInt32Number _cmsUtf16UnitsForScalar(cmsUInt32Number uc)
+{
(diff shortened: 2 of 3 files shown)


mm2:
Thanks for your contribution. **I am not accepting this PR for the following reasons**:

- Large portions of the code, especially those that appear AI-generated, require careful human review. Unless the underlying issue is important and such extensive changes are unavoidable, PRs of this size and nature will be rejected.  **The complexity and risk introduced by these changes do not justify the benefits**.

- The problem addressed is a corner case. In practice, most descriptions use characters from the Basic Multilingual Plane.

- LittleCMS is cross platform and the API must behave consistently across all systems. The proposed changes introduce behavior that only applies on platforms where wchar_t is 32-bit, 
  favoring it in the write direction. The current implementation intentionally limits itself to characters representable in 16 bits.

- To generate special characters a simple workaround exists: just use UTF-8.

- The PR introduces functionality that is **not part of the roadmap**, such as shared offsets and full support for supplementary multilingual planes.

I am, however, adding a small protective check so that bogus characters are converted to '?'.  49c1b0ff3c8c0185666e315fd768e26cdb6f985b Thanks for pointing out this issue.

PR #588, opened 2026-07-30
Remove generated testbed/new.icc from the tree

sallustfire:
I was hermetically building this and stumbled across testbed/new.icc, which is output of the test suite, not test data. The "Check MetaTag" test writes it fresh before ever reading it, and it's the only scratch profile the harness never cleans up with remove(). Being left behind after every run, it got committed by accident (a few times since 2022), and from there into the meson icc copy list, where it now breaks read-only checkouts:

```
$ chmod -R a-w Little-CMS
$ meson setup build Little-CMS && meson compile -C build && meson test -C build
...
Checking Check MetaTag ...
1/1 testcms FAIL   killed by signal 11 SIGSEGV
```

configure_file copies the corpus into the build dir preserving file modes, so the copied new.icc is read-only; fopen("new.icc", "wb") returns NULL and the unchecked fwrite crashes. A VPATH make check does the same, since cp also preserves the mode. As far as I can tell, writable checkouts never see any of this. There the test just silently rewrites the tracked file, which is why CI stays green and the only visible symptom has been the perpetually dirty tree.

This deletes the file, drops it from the meson list (configure_file errors on a missing input), and gitignores it. Verified: test output is byte-identical with the file absent, and the read-only meson run above passes after the change. The file isn't in EXTRA_DIST, so make dist tarballs never shipped it.

The leftover could also be eliminated at the source by having CheckMeta remove("new.icc") like every other test does with its scratch file. Then, nothing would ever leave the file behind, gitignored or not. Happy to add that here if you'd prefer; I left it out to keep this a pure deletion.


mm2:
Thanks for reporting. I am not adding this file to .gitignore since the issue comes from testcms2.c, so let's fix the root cause instead of ignoring the file. Fixed in d260167647910ca027fdd9b086dfcf1d95b2e6f8 and 1b1ec857842cd81e2d9b96ac576de35571209b42

sallustfire:
Thanks for the quick fix. Agreed that removing it rather than ignoring it a better approach.

PR #597, opened 2026-08-21
Fix ARM->x64 mapping

jaimergp:
Is this deliberate or an oversight?


mm2:
Ah, ok, this is undocumented, unsupported and AFAIK nobody is using it, so you couldn't break anything
Just joking, thanks for discovering the glitch and fixing it!

